PR #1, opened 2020-05-08
Some improvements

Awk34:


index.ts
```diff
@@ -1,7 +1,32 @@
 import {createLineChart} from './src/line';
 
-const chartDiv = document.getElementById('chart');
+if (module.hot) {
+  module.hot.accept();
+}
```
danielyule, line 5: Seems like you already have this logic down below?

Awk34:
wow, such thanks, many merge

PR #5, opened 2020-05-13
Deploy to gh-pages on every push to master

64json:


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+      - name: Deploy
+        uses: JamesIves/github-pages-deploy-action@releases/v3
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Awk34, line 26: why is this needed? How does this get populated?

64json: I think `JamesIves/github-pages-deploy-action` needs it to have permission to push to `gh-pages` branch. [`GITHUB_TOKEN`](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token) is provided by GitHub Actions when the workflow runs.

PR #6, opened 2020-05-13
switch back to relative paths in index.html

Awk34:


64json:
I tried building with this branch, but the paths in `index.html` resulted the same. I think it's Parcel replacing relative paths to absolute paths while compiling.

PR #7, opened 2020-05-13
Enforce Google TypeScript Style and Type Checking

64json:
- Use [google/gts](https://github.com/google/gts).
- Run `eslint --fix`.
- Fix typos.
- Run type check along with lint on every pull request ([as Parcel performs no type checking](https://parceljs.org/typeScript.html)).
- Fix type mismatches.


.eslintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./node_modules/gts/",
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "off"
```
Awk34, line 4: I think we'd want this on

64json: I agree but at the same time I was unsure what to put instead of `any` for the generic types of `d3.Selection<SVGGElement, any, any, any>`. I initially suppressed warning for that specific line but realized it will be used quite frequently.

64json: Should I create a type alias like below and re-enable the rule maybe?
```ts
// eslint-disable-next-line @typescript-eslint/no-explicit-any
type D3Selection<
  GElement extends d3.BaseType,
  Datum = any,
  PElement extends d3.BaseType = any,
  PDatum = any
> = d3.Selection<GElement, Datum, PElement, PDatum>;
```

src/line.ts
```diff
@@ -1,83 +1,93 @@
...
+  const svg = d3.create('svg').attr('viewBox', [0, 0, width, height].join(' '));
 
   const data = getData();
 
-  const xAxis = (g: d3.Selection<SVGGElement>) => g
-      .attr("transform", `translate(0,${height - margin.bottom})`)
-      .call(d3.axisBottom(x).ticks(width / 80).tickSizeOuter(0));
-  const yAxis = (g: d3.Selection<SVGGElement>) => g
-      .attr("transform", `translate(${margin.left},0)`)
+  const xAxis = (g: d3.Selection<SVGGElement, any, any, any>) =>
```
Awk34, line 40:
```suggestion
  const xAxis = (g: d3.Selection<SVGGElement, Datum[], SVGElement, Datum[]>) =>
```

64json: Per `@types/d3-selection`,
```
/**
 * A D3 Selection of elements.
 *
 * The first generic "GElement" refers to the type of the selected element(s).
 * The second generic "Datum" refers to the type of the datum of a selected element(s).
 * The third generic "PElement" refers to the type of the parent element(s) in the D3 selection.
 * The fourth generic "PDatum" refers to the type of the datum of the parent element(s).
 */
```

Shouldn't the 2nd and 4th generic still remain as `any` since the axis element and the svg element don't have a datum bound?

I've looked into [their test code](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/68c3c174697d1bf27af59a0c88d05a8d8eff454f/types/d3-selection/d3-selection-tests.ts#L765), and it seems like they also leave none-bound element as `any`.

64json: Actually, I think it makes more sense to have it as `d3.Selection<SVGGElement, undefined, SVGElement, undefined>)` since we never _define_ which datum to bind. How do you think?

64json: Or we can just inline `xAxis` and `yAxis` as in 7f1f9df to avoid explicitly using `any`. In this case, `g` on [line 72](https://github.com/googleinterns/guide-doge/pull/7/commits/7f1f9dfb4d921c111eedef970dc01d5ca09a5002#diff-2bcbe349d0874ef3d101f576c00dee35R72) is inferred to be `d3.Selection<SVGGElement, undefined, null, undefined>`. (I can revert the commit if it looks worse.)

.eslintrc.json
danielyule: We use something pretty similar to this internally for our d3 charts.  But instead of `any`, we use `unknown`.  In general, whenever `any` seems like a good idea, we should use `unknown` instead, unless an external library makes it impossible.

src/line.ts
```diff
@@ -1,83 +1,92 @@
...
-  return data.map((datum) => ({
-      date: startDate.plus({ days: datum.categories.get('nthDay') }),
-      value: datum.values.get('activeUsers')!,
+    ['nthDay'],
+    ['activeUsers'],
+    [betweenDates(startDate.toJSDate(), endDate.toJSDate())]
+  );
+  return data.map(datum => ({
+    date: startDate
+      .plus({days: datum.categories.get('nthDay') as number})
```
danielyule, line 25: Generally better to use the `!` operator here than a cast, because it makes it more clear that you're only saying the undefined value is guaranteed not to be there.

64json: Hi @danielyule, `days` requires to be of `number | undefined` type while `datum.categories.get('nthDay')` is of `string | number | undefined`, so I had to explicitly cast it to `number`.

.eslintrc.json
64json: Good call, I'll try using `unknown` next time.

PR #8, opened 2020-05-15
Encapsulation and minor refactoring

64json:


src/datagen/MockDataCube.ts
```diff
@@ -0,0 +1,88 @@
...
+ *  - The number of hits is normally distributed with mean settings.avgHits
+ *    and standard deviation settings.hitsStdDev.
+ *  - The value of each hit for measures of type SUM is uniformly distributed in
+ *    its range.
+ *  - The session a hit belongs to is uniformly distributed.
+ *  - The weight of a day is normally distributed with mean 1 and standard
+ *    deviation settings.dailyStdDev
+ *
+ */
+export class MockDataCube extends DataCube {
```
danielyule, line 49: I've a pretty strong preference to avoid inheritance when possible.  I understand that the idea here is that one could have other types of cubes that get data from sources like a csv file or online.  But the source of the data isn't an important property of the cube, and therefore shouldn't be included in its class.  That's why the original approach was using the [factory pattern](https://en.wikipedia.org/wiki/Factory_method_pattern )

Where the use of subclasses for construction can cause problems down the line is if you want to create a new type of cube called like `DataCubeWithMachineLearning` or something, then it's not clear what this new class should inherit from.  If it inherits from `MockDataCube` then any other data cubes won't get the new stuff.  If you inherit from `DataCube` then you won't be able to use it with Mock data.

Also, just in general, inheritance creates complexity that usually isn't needed.  Factorys and composition can usually handle the same needs, but in a more flexible and loosely coupled way.

src/datagen/types.ts
```diff
@@ -103,3 +122,8 @@ export interface ResultRow {
   categories: Map<string, string | number>;
   values: Map<string, number>;
 }
+
+export interface TrieNode {
+  children: Record<string, TrieNode>;
+  values?: number[];
+}
```
danielyule, line 131: Similar to above, this doesn't need to be defined out of the file that uses it.

src/datagen/utils.ts
```diff
@@ -121,7 +29,7 @@ function initalizeUsersAndSessions({
   return sessions;
 }
 
-function getNormalizedWeights(categories: Category[]) {
+export function getNormalizedWeights(categories: Category[]) {
```
danielyule, line 32: Internally, we have a rule which says that every exported function must have a JSDoc comment describing what it does.  By marking this with export, this method is now part of the public API of this module.

src/datagen/types.ts
```diff
@@ -47,15 +47,31 @@ export enum Scope {
...
 
+export interface User {
+  rowsSeen: Set<number>;
+}
+
+export interface Session {
+  user: User;
+  rowsSeen: Set<number>;
+}
+
```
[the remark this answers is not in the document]
64json: That makes sense, I will revert those changes.

src/datagen/MockDataCube.ts
64json: I see your point that the factory pattern fits better in this case. At the same time, I still think using `class` has quite a lot of advantages considering reusability and encapsulation.

In that sense, how do you think of having `DataCubeFactory` class instead? It would have:
- a constructor initializing factory settings
- public setter methods to manipulate factory settings
  - following the method chaining pattern of D3.js
- private helper methods
- `generate` public method

The primary advantage would be that users can create multiple data cubes with one factory setting without repeating code.

The usage would look like:
```ts
const factory = new DataCubeFactory(categories, measures, {
  avgHits: 10000,
  hitStdDev: 100,
  avgUsers: 100,
  userStdDev: 1,
  days: 30,
});
const monthlyData1 = factory.generate();
const monthlyData2 = factory.generate();
const annualData = factory
  .days(365)
  .generate();
const annualDataWithAdditionalCategory = factory
  .categories([...categories, anotherCategory]);
  .generate();
```

src/datagen/utils.ts
64json: Do you know if there is a ESLint rule for that? We are currently using [gts (Google TypeScript Style)](https://github.com/google/gts) and think it would be great if I can add the rule on top of it.

danielyule: It looks like there isn't a specific rule for this one.  I suspect the internal TS team wrote something themselves.

src/datagen/MockDataCube.ts
danielyule: I think what you're looking for might be closer to the [builder pattern](https://refactoring.guru/design-patterns/builder).

For encapsulation, you don't need a class; file level encapsulation is fine for a lot of uses, although not for sharing state.  In TS, files are modules, so everything in a file belongs to that module.

For re-usability, instead of returning a class, our factory can return a generator which is a function that can create a cube based on the current settings.  The filters code follows this pattern.   This approach does make it hard to change the settings later.

If you do want to go with the builder pattern (which given the use case from your example, you would), then using a class makes sense.  

With all that said, the design of data cube is that all visualizations on a page would use the same cube, so all this elaborate building may not be necessary.  In Google Analytics, all of the data across every report comes from the same cube.  Essentially, the visualizations (or in this case audifications) are showing the same data from many different angles.  This might be a good pattern here as well, but of course this is a demo app, not a re-implementation of GA.  I leave the final decision in your hands.

64json: Oh, I was thinking of the inheritance or factory/builder pattern because I thought we might create multiple data cubes, but I just realized that GA only needs one cube throughout the app indeed. I don't see the reason to not follow how GA uses a cube. I'll stick with your initial design. Thanks for letting me know!

64json:
> These are looking very nice.

Thank you!

PR #10, opened 2020-05-27
Line Chart Audification

64json:


src/d3/line-chart.d3.ts
```diff
@@ -1,27 +1,102 @@
...
+
+
+    const dataSubscription = dataObservable.subscribe(data => {
+      path
+        .datum(data)
+        .transition(this.transition)
+        .attr('d', line);
+    });
+
+    return () => {
```
jnager, line 36: Would it make sense to split up the rendering of data and the removal of the rendered data into two methods? Then the base chart could handle all pipes / subscriptions to the dataObservable and the destroy methods could be called on finalize() or something like that? I think it would reduce complexity if we could limit the subscriptions to the dataObservable.

src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -1,25 +1,96 @@
...
+import { Melody } from '../../models/melody/melody.model';
 
 @Component({
   selector: 'app-line-chart-audification',
   templateUrl: './line-chart-audification.component.html',
 })
-export class LineChartAudificationComponent {
-  public disposeMelody?: () => void;
+export class LineChartAudificationComponent implements OnInit, OnDestroy {
+  @Input() frequencyRange: [number, number] = [256, 2048];
```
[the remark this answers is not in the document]
64json: I arbitrarily chose the ones that sound most comfortable to me personally among powers of 2 (for no reason 😅). I am planning on letting users set those values as they prefer.

src/components/line-chart-audification/line-chart-audification.component.html
```diff
@@ -1,3 +0,0 @@
-<div role="button" (focus)="playMelody()" (blur)="disposeMelody && disposeMelody()">
```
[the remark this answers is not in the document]
64json: There will be a template/style to this component. (In #14: [line-chart-audification.component.html](https://github.com/googleinterns/guide-doge/blob/16a6f9c17c1624c219001a0f7bb425e95f5a89ea/src/components/line-chart-audification/line-chart-audification.component.html)) Though, I agree that it would semantically look better with a directive if a template was not needed.

src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -1,25 +1,96 @@
...
+      window.clearTimeout(this.refocusTimeout);
+    }
+  }
+
+  async handleKeyDown($event: KeyboardEvent) {
+    const { key, shiftKey, repeat } = $event;
+    if (repeat) {
+      return;
+    }
+    if (key === ' ') {
```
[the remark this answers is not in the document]
64json: I don't think I'll be adding more shortcuts for the same action in order to minimize conflicts with screen-reader applications. (Speaking of which, it might be a good idea to let users customize shortcuts.) I'll go with `if` for now, but I'll consider using `switch ... case` when I implement the shortcut customization.

src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -1,25 +1,96 @@
...
+      this.melody = new Melody(values, this.frequencyRange, this.duration, (index, playing) => {
+        this.component.activeDatum = data[index];
+        if (!playing) {
+          this.readOut(this.lineChartD3.activeGroupId);
+        }
+      });
+    });
+    this.element.addEventListener('keydown', this.handleKeyDown);
+    this.element.addEventListener('keyup', this.handleKeyUp);
+  }
```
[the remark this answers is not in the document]
64json: Didn't know about and that sounds better!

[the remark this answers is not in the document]
64json: Actually, while looking `fromEvent` up, I learned I can just use `@HostListener`, which makes it possible to avoid using `elementRef` and manual unsubscribing.

src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -1,25 +1,96 @@
...
   ) {
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleKeyUp = this.handleKeyUp.bind(this);
   }
 
-  async playMelody() {
-    if (Tone.getContext().state === 'suspended') {
-      await Tone.start();
+  ngOnInit() {
+    this.dataSubscription = this.component.dataObservable.subscribe(data => {
```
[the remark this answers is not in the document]
64json: I love this pattern, thanks!

src/d3/line-chart.d3.ts
64json: It does look much less complex after splitting them up, thanks! Indeed, there is nothing much to tear down since utilizing the destroy$ pattern.

src/models/melody/melody.model.ts
```diff
@@ -0,0 +1,134 @@
...
+    const sequence = new Tone.Sequence(
+      (time, value) => {
+        this.seekTo(this.currentSeconds);
+        const keyNumber = (value - minValue) / (maxValue - minValue) * (maxKeyNumber - minKeyNumber) + minKeyNumber;
+        const frequency = Melody.getFrequency(keyNumber);
+        this.synth.triggerAttackRelease(frequency, '4n', time);
+      },
+      values,
+    );
+    sequence.playbackRate = values.length / 4 / this.duration;
```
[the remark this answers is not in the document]
64json: Each note takes 1/4 seconds, hence the total duration is `values.length * 1 / 4`. I set the playback rate to the total duration divided by the desired duration (`this.duration`).

I also find it overly complicated, so I refactored it in a way that the duration of each note is adjusted but not the playback rate.

src/models/melody/melody.model.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  getCurrentDatumIndex() {
+    return this.currentDatumIndex;
+  }
+
+  seekTo(seconds: number, inclusive = false) {
+    this.currentDatumIndex = this.getDatumIndex(seconds);
+    this.inclusive = this.isEnded || inclusive;
+    this.onSeek?.(this.currentDatumIndex, this.isPlaying);
```
jnager, line 97: Is this callback necessary? Could the LineChartAudification component instead just get the currentDatumIndex and isPlaying from this.melody and execute its logic in order to avoid the need to pass the callback down?

src/d3/xy-chart.d3.ts
```diff
@@ -1,78 +1,76 @@
...
-      );
-
-    this.appendChart(svg, data, scaleX, scaleY);
-
-    return () => {
-      svg.remove();
-    };
+      .attr('transform', `translate(${marginLeft},0)`);
+
+    dataObservable
```
jnager, line 54: I'm probably being a bit picky here, but I think it might be easier to understand the render flow if this method contains the only subscribe to the dataObservable. Here's what I imagine it would look like:

```
render() {
  this.dataObservable
    .pipe(this.takeUntilCleared())
    .subscribe(data => {
      this.renderAxis(data);
      this.renderData(data);
      this.renderActiveDatum(data);
    });

  this.activeDataObservable
    .pipe(this.takeUntilCleared())
    .subscribe(datum => {
      this.renderActiveDatum(datum)
      // Could contain other actions based on activeDatum
    });
}
```

Tracking one subscription (or one for each observable) seems cleaner and potentially less error-prone to me. Also, would it be easier to store the RenderOptions as an instance variable on the base class, rather than passing them through these methods?

WDYT?

src/components/line-chart-visualization/line-chart-visualization.component.scss
```diff
@@ -0,0 +1,7 @@
+:host {
+  ::ng-deep {
```
jnager, line 2: What about just adding a container element in the line-chart component and setting a max-width on that?

src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -1,25 +1,116 @@
...
-    private audificationService: AudificationService,
+    private zone: NgZone,
   ) {
+    this.handleSeek = this.handleSeek.bind(this);
   }
 
-  async playMelody() {
-    if (Tone.getContext().state === 'suspended') {
-      await Tone.start();
+  get INSTRUCTIONS() {
```
jnager, line 29: I'm a bit torn on using the custom i18n solution inside an Angular component, since Angular has their own i18n features.
https://angular.io/guide/i18n

I really like the i18n custom attribute and find it makes templates super easy to understand. It would be nice to have the text displayed in the HTML and not have to track down AUDIFICATION.INSTRUCTIONS to see what it actually says.

With that said, using two different systems of internationalization increases the translation complexity, so it makes sense to continue to use the custom i18n translate method used in the D3 files for this project.

Basically - I guess this wasn't a comment for change so much as me just musing aloud. :)

src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -1,25 +1,116 @@
...
+  private melody?: Melody;
+  private domain: Date[];
+  private range: number[];
 
   constructor(
-    @Host() public component: LineChartComponent,
-    private audificationService: AudificationService,
+    private zone: NgZone,
   ) {
+    this.handleSeek = this.handleSeek.bind(this);
```
Awk34, line 26: You could instead use the [autobind decorator](https://github.com/jayphelps/core-decorators#autobind)

src/models/melody/melody.model.ts
```diff
@@ -0,0 +1,134 @@
...
+    return sequence;
+  }
+
+  private getSeconds(index: number) {
+    return (index + .5) * this.noteDuration;
+  }
+
+  private getDatumIndex(seconds: number) {
+    const index = Math.round(seconds / this.noteDuration - .5);
+    return Math.min(Math.max(index, 0), this.values.length - 1);
```
Awk34, line 132: Is this correct? It seems like the `min` & `max` calls are backwards

64json: I think it is correct.
When `index = -1` or less, it will be adjusted to `0`.
When `index = this.values.length` or greater, it will be adjusted to `this.values.length - 1`.

src/models/melody/melody.model.ts
64json: I needed to pass the callback in order for the component to know **when** to update the active datum on the visualization and to read it out.

src/components/line-chart-audification/line-chart-audification.component.ts
64json: I see the pros and cons of both methods. Do you know how GA is handling i18n in d3 manipulation?

PR #11, opened 2020-05-27
I18n

64json:
Internationalization is not of priority in this repo, but it indirectly helps manage hard-coded texts in one place.


src/assets/i18n/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+  return language;
+}
+
+/**
+ * Returns a formatted i18n string using sprintf-js.
+ *
+ * @param key The key of an i18n string.
+ * @param args The arguments to format the string with.
+ */
+export function t(key: keyof I18n, ...args: any[]) {
```
jnager, line 25: Is this "t" for "translate"? If so, perhaps just use the full word for ease of reading?

src/assets/i18n/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface I18n {
```
jnager, line 1: What about creating a string enum with the values you always want populated for each language and then making I18n a AudificationLabelEnum (or some better name) : string type? Then you could use the enum to select values in the d3 and line-chart code below.

src/assets/i18n/utils.ts
64json: Yes, and I brought the idea from [react-i18next](https://github.com/i18next/react-i18next). Since this function would be used very often in both .html and .ts, I think it would rather improve the overall readability if its name is as short as possible.

src/assets/i18n/types.ts
64json: Great idea! I'll try that way.

PR #12, opened 2020-05-28
Unit testing

64json:


src/models/melody/melody.model.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { Melody } from './melody.model';
+import * as Tone from 'tone';
+
+describe('Melody', () => {
+  const values = [1, 2, 3, 4, 5, 4, 3, 2, 1];
+  const frequencyRange: [number, number] = [256, 2048];
+  const duration = 5;
+  const delay = 200;
```
jnager, line 8: Could you add comments explaining why you picked the duration and delay amounts here? (Or just indicate the units.)

PR #14, opened 2020-06-05
GA-like dashboard layout

64json:
Screen Shot 2020-06-05 at 5 55 32 PM


src/components/fold/fold.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
```
jnager, line 3: What is the FoldComponent doing?

src/components/switch/switch.component.ts
```diff
@@ -0,0 +1,17 @@
+import { Component, EventEmitter, Input, Output } from '@angular/core';
```
jnager, line 1: Purely FYI - in our own codebase, we have been trying more and more to move to the Angular Material components (mostly) for these kinds of presentational components. This helps keep styling consistent and reduces our code maintenance costs. So for this, we would use:
https://material.angular.io/components/slide-toggle/overview

Similarly, there are Angular Material cards, tabs, and expansion panels that could be used to replicate other aspects of the dashboard functionality built here.

With that said, writing your own components seems like more fun for this project so the SwitchComponent seems great! Just wanted to share our own practices and let you know about Angular Material in case you didn't know about that option / wanted to try it out.

src/components/line-chart/line-chart.component.html
```diff
@@ -1,8 +1,20 @@
-<button (click)="toggleMeasure()">Toggle Measure</button>
-<app-line-chart-visualization role="img"
-                              [attr.aria-label]="VISUALIZATION"
-                              [data]="data"
-                              [(activeDatum)]="activeDatum"></app-line-chart-visualization>
-<app-line-chart-audification tabindex="0"
-                             [data]="data"
-                             [(activeDatum)]="activeDatum"></app-line-chart-audification>
+<app-card class="card" title="Line Chart">
```
jnager, line 1: Should app-line-chart be a child of app-card (instead of vice versa)? It seems like the cards contain the visualizations.

src/components/fold/fold.component.ts
64json: It is a collapsible wrapper component of each section (audification / visualization) within a card that adds divider and padding. I'll rename it to `CardFoldComponent` for clarity.

src/components/line-chart/line-chart.component.html
64json: That makes sense. Would it be okay if I rename the component to `line-chart-card` instead?

src/app.component.scss
```diff
@@ -0,0 +1,77 @@
+@import "./utils/constants";
+
+:host {
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+  height: 100vh;
+  background-color: $color-background;
+
+  .header {
```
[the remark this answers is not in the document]
64json: Is it because of the concern of possibly duplicating properties?

I personally like to group them with similar properties (dimensions, flex-related, text-related, etc.) since modern IDEs warn the duplicates. I feel like `width` not with `height`, or `align-items` not with `justify-content` would look a bit off.

src/components/switch/switch.component.ts
64json: I tried to DIY for learning purposes to understand the data flow of Angular better (since it is my first Angular project that I wrote from scratch), but I definitely agree that using Angular Material would make the project focus more on logic.

PR #15, opened 2020-06-10
Fix broken gh-pages deployment

64json:
I found a bug that deployment commits pushed to `gh-pages` are not actually reflected on https://googleinterns.github.io/guide-doge.

Per the documentation of [GitHub Pages Deploy Action](https://github.com/marketplace/actions/deploy-to-github-pages#configuration-):
> If you experience any issues with your changes not being reflected after the deployment it may be neccersary to use either the SSH or ACCESS_TOKEN options.

I believe this is our case, so I changed the GH Action to use the SSH option instead of GITHUB_TOKEN.

**NOTE** that before this pull request is merged, one of the repo admins needs to follow [this instructions](https://github.com/JamesIves/github-pages-deploy-action/tree/dev#using-an-ssh-deploy-key-).


64json:
@Awk34 Thanks in advance :)

Awk34:
I've added an SSH deploy key as DEPLOY_KEY in the repo's secrets. Follow https://github.com/JamesIves/github-pages-deploy-action/tree/dev#using-an-ssh-deploy-key- to use it.

PR #16, opened 2020-06-10
Read out legend items on pressing L

64json:
(branched off of #14)


src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -84,6 +85,8 @@ export class LineChartAudificationComponent implements OnDestroy, OnChanges {
         min: formatY(this.range[0]),
         max: formatY(this.range[this.range.length - 1]),
       }));
+    } else if (key === 'l') {
+      this.readOut(humanizeMeasureName(this.measureName));
```
jnager, line 89: What about including the x axis? E.g. "active users by day" or "active users each month"?

64json: I was thinking of including it when pressing <kbd>X</kbd>: `Every day from May 4th to August 7th` because later when there are multiple legend items, it would be redundant to say `by day` for each legend item.

PR #21, opened 2020-06-12
Component hierarchy

64json:
This PR mainly deals with:
1. Revise component hierarchy similar to GA.
2. Create `audification` directive to dynamically attach audification components.


src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -35,18 +36,34 @@ export class LineChartAudificationComponent implements OnDestroy, OnChanges {
...
+  set activeDatum(activeDatum) {
+    this.host.activeDatum = activeDatum;
+  }
+
+  ngOnInit() {
+    this.host.data$
+      .pipe(takeUntil(this.destroy$))
+      .subscribe(data => {
+        const values = data.map(datum => datum.value);
+        this.domain = data.map(d => d.date).sort((a, b) => a.getTime() - b.getTime());
```
Awk34, line 56: nit: can this sort function be pulled out into a pure util function?

src/directives/audification/audification.directive.ts
```diff
@@ -0,0 +1,48 @@
...
+
+@Directive({
+  selector: '[appAudification]',
+})
+export class AudificationDirective implements OnInit, OnDestroy {
+  audificationComponentRef: ComponentRef<unknown> | null;
+
+  constructor(
+    // candidate host components
+    @Host() @Self() @Optional() private lineChartComponent: LineChartComponent | null,
```
Awk34, line 13: Can we have a more scale-able solution to this?

src/components/line-chart-audification/line-chart-audification.component.ts
64json: Sounds good. I also just realized that the line below that
```ts
this.range = this.data.map(d => d.value).sort();
```
has been wrongly sorting in alphabetical order instead of in numerical order.

src/directives/audification/audification.directive.ts
64json: I tried to use Union Types, but Angular gave me JIT compilation error.

I also tried to make an abstract parent class, let `LineChartComponent` and `BarChartComponent` extend it, and change the signature to `@Host() @Self() private hostComponent: ParentClass | null`, but `@Host()` injected `null` to it.

Let me know if you have other solutions that I can try.

PR #22, opened 2020-06-16
User preferences

64json:
(Branched off of #21.)

- Created `preference-group.component` and `preference-item.component`.
- Created `preference.service` that has each preference as an Observable and provides a combined Observable for each preference group.
- Modified `line-chart-audification.component` and `audification.directive` to reflect the user preferences.

There is a bug that the active indicator does not synchronize with the melody well when a note duration is set to be short enough. -> It will be addressed in a separate PR soon.


src/components/preference-item/preference-item.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, Input } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
+
+@Component({
+  selector: 'app-preference-item',
+  templateUrl: './preference-item.component.html',
+  styleUrls: ['./preference-item.component.scss'],
+})
+export class PreferenceItemComponent<T> {
```
Awk34, line 9: this generic seems pointless as it isn't set anywhere

src/services/preference/preference.service.ts
```diff
@@ -0,0 +1,54 @@
...
+
+type ObservableDictionary = { [key: string]: Observable<any> };
+type ObservedDictionaryOf<T extends ObservableDictionary> = { [key in keyof T]: ObservedValueOf<T[key]> };
+
+export class PreferenceService {
+  audification = { // mimic namespace
+    enabled: new BehaviorSubject(true),
+    lowestPitch: new BehaviorSubject(256),
+    highestPitch: new BehaviorSubject(1024),
+    noteDuration: new BehaviorSubject(.167),
```
Awk34, line 13: should be `const` values

src/components/preference-item/preference-item.component.ts
64json: I think it would be less error prone to use generics (even though it isn't set anywhere) than making the type of `subject` to `BehaviorSubject<any>` or `BehaviorSubject<number | boolean | string>`.

```ts
subject: BehaviorSubject<T>;

// compile-time errors, which prevent runtime errors
subject.next('some_string');
subject.next(42);
```

```ts
subject: BehaviorSubject<any>;

// no compile-time errors, but potential runtime errors
subject.next('some_string');
subject.next(42);
```

src/services/preference/preference.service.ts
64json: you mean making the properties non-observable, or replacing the hard-coded default values with `const` variables?

Issue #23, opened 2020-06-16
The active indicator does not synchronize well.


64json:
The active indicator does not synchronize well with the melody when a note duration is set to be short enough.

PR #24, opened 2020-06-18
Fix issue #20 and #23

### LINKED ISSUE #20: Melody should stop Tone.Transport upon dispose
https://github.com/googleinterns/guide-doge/blob/f01fa634a0e493596aa578cb3a0124fba73ecba1/src/models/melody/melody.model.spec.ts#L21
Move line 21 **into** the `dispose()` method of `Melody`.

64json:
(Branched off of #22.)
Use `window.setTimeout` instead of `Tone.Transport` to play melody. Even though it is not accurate to within milliseconds, it guarantees to never skip a note and synchronize the active indicator with playhead perfectly.

### DIFF
--- a/src/components/dashboard/dashboard.component.html
+++ b/src/components/dashboard/dashboard.component.html
@@ -3,7 +3,7 @@
   <app-preference-group title="Audification" [subject]="audification.enabled">
     <app-preference-item name="Lowest note (Hz)" [subject]="audification.lowestPitch"></app-preference-item>
     <app-preference-item name="Highest note (Hz)" [subject]="audification.highestPitch"></app-preference-item>
-    <app-preference-item name="Note duration (sec)" [subject]="audification.noteDuration"></app-preference-item>
+    <app-preference-item name="Note duration (ms)" [subject]="audification.noteDuration"></app-preference-item>
     <app-preference-item name="Read out before playing" [subject]="audification.readBefore"></app-preference-item>
     <app-preference-item name="Read out after playing" [subject]="audification.readAfter"></app-preference-item>
   </app-preference-group>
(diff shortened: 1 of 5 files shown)


src/components/preference-group/preference-group.component.ts
```diff
@@ -0,0 +1,20 @@
...
+import { BehaviorSubject } from 'rxjs';
+
+@Component({
+  selector: 'app-preference-group',
+  templateUrl: './preference-group.component.html',
+  styleUrls: ['./preference-group.component.scss'],
+})
+export class PreferenceGroupComponent {
+  @Input() title: string;
+  @Input() subject: BehaviorSubject<boolean>;
```
jnager, line 11: How about preference$ or some other name with a bit more of an indication of what value the user can expect?

src/components/preference-item/preference-item.component.ts
```diff
@@ -0,0 +1,24 @@
...
+import { BehaviorSubject } from 'rxjs';
+
+@Component({
+  selector: 'app-preference-item',
+  templateUrl: './preference-item.component.html',
+  styleUrls: ['./preference-item.component.scss'],
+})
+export class PreferenceItemComponent<T> {
+  @Input() name: string;
+  @Input() subject: BehaviorSubject<T>;
```
jnager, line 11: ditto here for the naming

src/components/preference-group/preference-group.component.ts
64json: sounds good! one thing i'm worried about is is it conventionally okay to have `$` in the name of an input property?
```html
<app-preference-group [preference$]="...">
```

PR #25, opened 2020-06-18
Make CI fail when test coverage is under 85%

64json:


karma.conf.js
```diff
@@ -19,6 +19,15 @@ module.exports = function (config) {
       dir: require('path').join(__dirname, './coverage/guide-doge'),
       reports: ['html', 'lcovonly', 'text-summary'],
       fixWebpackSourcePaths: true,
+      thresholds: {
+        emitWarning: false,
+        global: {
+          statements: 85,
```
Awk34, line 25: any reason 85? Can we warn on a higher number and fail on a slightly lower number?

64json: I love the idea, but it seems karma only lets me either emit error all the time or emit warning all the time.
https://www.npmjs.com/package/karma-coverage-istanbul-reporter

PR #27, opened 2020-06-21
Add tests for components

64json:
(Branched off of #24)


src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -24,14 +24,14 @@ export class LineChartAudificationComponent implements AudificationPreference, O
...
   private domain: Date[];
   private range: number[];
   @HostBinding('attr.tabindex') private readonly tabindex = 0;
   private readOutTimeoutId: number | null = null;
 
   constructor(
-    @Inject('host') private host: LineChartComponent,
+    @Inject('host') public host: LineChartComponent,
     private zone: NgZone,
   ) {
```
jnager, line 36: We try to avoid making class items public purely for the sake of testing. Is there another way to test this component without accessing private instance variables and methods? 

Could you create an instance of the host lineChartComponent and supply that instance using `useValue` in the providers array? And then test the readOut method through simulating keyUp and keyDown events?

src/components/line-chart/line-chart.component.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(component).toBeInstanceOf(LineChartComponent);
+  });
+
+  it('should have truthy i18n values.', () => {
+    component.activeDatum = mockDatum;
+    expect(component.ACTIVE_DATUM).toBeTruthy();
+  });
+
+  it('should synchronize data with data$.', () => {
+    component.data = [];
```
jnager, line 34: We have been trying to write new tests using the Arrange - Act - Assert structure. So all the set up work is done in one block at the start of the test, followed by some actions, followed by assertions. It has helped us to keep our tests a bit cleaner. What do you think about using that pattern here?

src/components/line-chart-audification/line-chart-audification.component.ts
64json: That makes sense. I was able to make `host` private using `useValue`.

For the `readOut` method, I moved it out to `ScreenReaderComponent` and made it public in #34.

src/components/line-chart/line-chart.component.spec.ts
64json: I see the pattern can make some tests cleaner, but wouldn't it be less straightforward for the tests where the order of `expect` statements matters? If I were to follow the pattern in this test, it adds unnecessary lines to store the value of the variables after each 'Act'.

```ts
it('should synchronize data with data$.', () => {
  component.data = [];
  const observedValue1 = component.data$.value;
  const value1 = component.data;

  component.data = [mockDatum];
  const observedValue2 = component.data$.value;
  const value2 = component.data;

  expect(observedValue1).toBe(value1);
  expect(observedValue2).toBe(value2);
});
```

jnager: Do you need to test `observedValue2`? It seems to me that testing one instance of setting data should be sufficient. Or if not, you could break the test into two. I.e.`it('should set data as data$`, ...)` and 'it('should update data when data$ changes', ...)`.

PR #28, opened 2020-06-21
Add tests for directives

64json:
(Branched off of #24)


src/directives/audification/audification.directive.ts
```diff
@@ -14,7 +14,7 @@ export class AudificationDirective implements OnInit, OnDestroy {
   private destroy$ = new Subject();
 
   constructor(
-    private preferenceService: PreferenceService,
```
jnager, line 17: Same comment here about keeping class items private when possible (i.e. not public for testing). Can you provide a Spy version of the PreferenceService in the TestBed providers setup? Then this should remain private.

PR #29, opened 2020-06-21
Add tests for utils

64json:
(Branched off of #24)


src/utils/formatters.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+      expect(formatX(firstDayOfYear)).toBe('January 01');
+    });
+  });
+
+  describe('formatY', () => {
+    it('should format a whole number correctly.', () => {
+      expect(formatY(42)).toBe('42');
+    });
+
+    it('should format a decimal number correctly.', () => {
```
jnager, line 17: nit: could you specify what "correct" means here? I.e. round to one decimal place, if that is what it does.

PR #30, opened 2020-06-21
Add tests for services

64json:
(Branched off of #24)


src/services/preference/preference.service.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  it('should instantiate.', () => {
+    expect(preferenceService).toBeInstanceOf(PreferenceService);
+  });
+
+  it('should have combined observable preferences.', done => {
+    expect(preferenceService.audification$).toBeInstanceOf(Observable);
+    preferenceService.audification$.subscribe(audificationPreference => {
+      Object.values(audificationPreference).forEach(value => {
+        expect(value).not.toBeInstanceOf(Observable);
```
jnager, line 19: Might be helpful for future readers to quickly understand this method if there is also a simple assertion that audificationPreference equals an object of:
```
{
  enabled: true,
  lowestPitch: 256,
  ...
}
```
etc.

64json: I changed the structure of `PreferenceService` quite a bit in #35, so I'll address this when I update the tests.

PR #31, opened 2020-06-21
Refactor i18n

64json:
- Move `i18n` out of `assets` because Angular copies the entire `assets` directory when building
- Support Korean (second language needed for testing purposes)
- Add tests for i18n


64json:
> Very cool - thanks Jinseo!

I thank you!

PR #37, opened 2020-06-24
Add data generator preferences

jcc37:
- UI changes: added data generator control panel to the sidebar
- Added data preference to `PreferenceService`
- Modified `DataService` to provide data subject and reflect data preference


src/services/preference/types.ts
```diff
@@ -10,6 +10,15 @@ export interface AudificationPreference extends Preference {
   readAfter: boolean;
 }
 
+export interface DataPreference extends Preference {
```
64json, line 13: I think you can make it `export type DataPreference = ModelSettings & Preference`?

src/services/data/data.service.ts
```diff
@@ -7,39 +7,72 @@ import {
...
+    DataService.categories,
+    DataService.measures,
+    DataService.settings,
+  ));
+  private destroy$ = new Subject();
+
+  constructor(private preferenceService: PreferenceService) {
+    this.preferenceService.data$
+      .pipe(takeUntil(this.destroy$))
+      .pipe(throttleTime(1000))
```
64json, line 40: I think `throttleTime` pipe can be modified a bit to remove kinda redundant initialization of `dataCube$` in line 30.
https://stackoverflow.com/questions/37146081/rxjs-throttle-behavior-get-first-value-immediately

In that case, we can also remove `this.settings`.

src/services/preference/types.ts
jcc37: Actually `DataPreference` is not `DataCube`'s setting. It contains parameters you want users to try to create different datasets. I use a subset of `ModelSettings` just to demonstrate the playground works properly, and more meaningful preference items will be added in future PRs.

src/services/data/data.service.ts
jcc37: I replaced `BehaviorSubject` with `ReplaySubject` to remove redundant initialization, thanks!

64json: I just realized it should be `deboundTime`, not `throttleTime`.

jcc37: `throttleTime` with settings `{ leading: true, trailing: true}` works as expected

src/services/data/data.service.ts
```diff
@@ -50,29 +50,43 @@ export class DataService implements OnDestroy {
...
     this.destroy$.complete();
   }
+
+  observeData(queryOptions: DataQueryOptions) {
+    const {
+      categoryNames$ = new BehaviorSubject([]),
+      measureNames$ = new BehaviorSubject([]),
+      filters$ = new BehaviorSubject(undefined),
+      sortBy$ = new BehaviorSubject(undefined),
+    } = queryOptions;
```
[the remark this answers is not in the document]
64json: why haven't I thought that way?!!

src/i18n/types.ts
```diff
@@ -15,8 +15,18 @@ export enum AUDIFICATION {
...
+  avgHits,
+  hitStdDev,
+  avgUsers,
+  userStdDev,
+  avgSessionsPerUser,
+  sessionsPerUserStdDev,
+}
+
+export enum AUDIFICATION_PREFERENCE {
+  enabled = 0x400,
```
Rhombus33, line 29: just a question, how is this value used?

src/models/data-cube/filters.ts
```diff
@@ -1,35 +1,27 @@
...
-  const { excludeStartDate, excludeEndDate } = {
-    ...defaultOptions,
-    ...options,
-  };
+export function inOneOfDateRanges(dateRanges: [Date, Date][], rangeOptions: RangeOptions = {}): Filter {
+  const {
+    excludeStart = false,
+    excludeEnd = false,
+  } = rangeOptions;
+  const startOffset = excludeStart ? 1 : 0;
```
Rhombus33, line 13: Could we use a constant for the offset value "1"? To make it clear what time increment it is.

src/services/data/data.service.ts
```diff
@@ -7,40 +6,75 @@ import {
...
-  getMeasureOverDays(measureName: string, days = 30): Datum[] {
-    const categoryName = 'nthDay';
-    const endDate = DateTime.local();
-    const startDate = endDate.minus({ day: days });
+  ngOnDestroy() {
+    this.destroy$.next();
+    this.destroy$.complete();
+  }
+
+  observeData(queryOptions$: Observable<QueryOptions>) {
```
Rhombus33, line 54: Should be private if that used external to this service.

src/models/data-cube/generation.ts
```diff
@@ -91,8 +93,9 @@ const defaultSettings: ModelSettings = {
   userStdDev: 100,
   avgSessionsPerUser: 5,
   sessionsPerUserStdDev: 3,
-  nthDay: true,
-  days: 60,
+  timeSeries: true,
+  startDate: new Date(Date.now() - 60 * DAY),
```
Rhombus33, line 97: Have constant for 60

src/services/data/data.service.ts
```diff
@@ -7,40 +6,75 @@ import {
...
-        [categoryName],
-        [measureName],
-        [betweenDates(startDate.toJSDate(), endDate.toJSDate(), { excludeStartDate: true })],
-      )
-      .map(row => ({
-        date: startDate
-          .plus({ days: row.categories.get(categoryName) as number })
-          .toJSDate(),
-        value: row.values.get(measureName)!,
+  observeTimeSeries(timeSeriesQueryOptions$: Observable<TimeSeriesQueryOptions>) {
```
Rhombus33, line 61: Add documentation that this ultimately returns data? or update method name to reflect that?

src/components/line-chart/line-chart.component.ts
```diff
@@ -15,16 +18,20 @@ import { DataService } from '../../services/data/data.service';
 export class LineChartComponent implements RenderOptions, OnChanges, OnInit, OnDestroy {
   @ViewChild(A11yPlaceholderDirective, { static: true }) a11yPlaceholder: A11yPlaceholderDirective<LineChartComponent>;
 
-  @Input() measureName: string;
+  @Input() endDate = new Date();
+  @Input() startDate = new Date(this.endDate.getTime() - 30 * DAY);
+  @Input() measureNames: string[] = [];
```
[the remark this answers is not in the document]
jcc37: Yes, they are metric names. "Measure" and "Category" are terms used in `DataCube`, and the input here takes the measure names of data provided by `DataService`.

src/models/data-cube/data-cube.model.ts
```diff
@@ -113,17 +113,18 @@ export class DataCube {
...
    * the second, and so on.
    *
-   * @param categoryNames The categories to request a breakdown from.
-   * @param measureNames The measures to provide values for.
-   * @param filters The filters to apply to the cube before finding the results.
-   * @param sortBy The concept names to sort in ascending order.
+   * @param queryOptions.categoryNames The categories to request a breakdown from.
+   * @param queryOptions.measureNames The measures to provide values for.
+   * @param queryOptions.filters The filters to apply to the cube before finding the results.
+   * @param queryOptions.sortBy The concept names to sort in ascending order.
```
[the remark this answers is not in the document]
jcc37: @64json I think we can move them to where the interface is defined.

src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -165,7 +165,8 @@ export class LineChartAudificationComponent implements AudificationPreference, O
   }
 
   private readOutMeasure() {
-    return this.screenReaderComponent.readOut(humanizeMeasureName(this.measureName));
+    // TODO: read the entire measureNames
+    return this.screenReaderComponent.readOut(humanizeMeasureName(this.measureNames[0]));
```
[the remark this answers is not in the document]
jcc37: No. `measureNames` here is hard-coded in `dashboard/dashboard.component.html`

src/models/data-cube/data-cube.model.ts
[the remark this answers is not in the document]
64json: Yeah that sounds better

src/i18n/languages/en.ts
```diff
@@ -25,6 +26,14 @@ export const en: I18n = {
   [AUDIFICATION.RANGE]: 'Range from %(min)s to %(max)s',
   [AUDIFICATION.ACTIVE_DATUM]: '%(y)s on %(x)s',
 
+  [DATA_PREFERENCE.enabled]: 'Data Generator',
+  [DATA_PREFERENCE.avgHits]: 'Average Hits',
+  [DATA_PREFERENCE.hitStdDev]: 'Hit Std',
```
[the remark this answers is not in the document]
jcc37: @64json We can modify it to 'Hit Standard Deviation`

src/i18n/types.ts
jcc37: `I18nKey` is a union type of all these enums. These values are used to resolve the conflict when defining i18n text for these enums in the same object.

src/services/data/data.service.ts
64json: This method has been completely moved and renamed in #46 and its child branches. I'll address that in those!

src/services/data/data.service.ts
jcc37: This part (data retrieval and `DataService`) is thoroughly modified and cleaner in #46 .

src/i18n/languages/en.ts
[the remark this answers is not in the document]
jcc37: @Rhombus33 If I change all "Std" to "Standard Deviation", the text for configurable items in the sidebar would wrap, which makes it ugly. I am thinking to add aria-label for screen readers and keep short for display, but not sure how people deal with this problem. What do you think?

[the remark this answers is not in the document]
Rhombus33: Having a separate aria label SGTM

src/i18n/types.ts
Rhombus33: Could you just add a comment describing what these values are for? Thanks!

Issue #38, opened 2020-06-24
Numerical preference item does not work when input is cleared


jcc37:
The input field in `preference-item` component disappears and never appears again until refreshing page when the value is cleared.

64json:
good catch! In addition to that, the a11y components would keep getting rerendered while typing numbers, so we should consider only updating when pressing <kbd>ENTER</kbd> or losing focus from `<input>`.

jcc37:
@64json Isn't that a feature? My workaround is to pipe the preference subject to throttle to prevent the rapid change of the value.

64json:
> @64json Isn't that a feature? My workaround is to pipe the preference subject to throttle to prevent the rapid change of the value.

That sounds better.

64json:
Closed by #42

PR #42, opened 2020-06-25
Fix issue #38

jcc37:
(Branched off of #37 )

Close issue #38 
Not push the input value to preference service when it is invalid e.g. null when the input is cleared.


jcc37:
> Just curious, is U a common shorthand for property name or key?

Actually P (property) and K (key) are used more often in TS documentation and Angular. I will modify it, thanks!

PR #43, opened 2020-06-26
Add rolling measure and period-over-period measure

64json:
Add methods in `DataService` to populate rolling measure and period-over-period measure after querying to `DataCube`.

Rolling measure would be to represent:
Screen Shot 2020-06-26 at 6 02 58 PM

Period over period measure:
Screen Shot 2020-06-26 at 6 04 54 PM


64json:
Closing it, as the core changes in this PR have been moved to another file in #46.

PR #44, opened 2020-06-26
Apply material design

64json:


src/components/card/card.component.html
```diff
@@ -1,5 +1,5 @@
 <h2 class="card-title">{{title}}</h2>
-<div class="card-content">
+<mat-card class="card-content">
   <div class="tabs" *ngIf="tabbed">
```
jnager, line 3: You could switch this to `mat-tab`s too, if you want to.

src/components/preference-group/preference-group.component.html
```diff
@@ -1,7 +1,9 @@
-<label class="header" [class.enabled]="enabled" [class.always-enabled]="alwaysEnabled">
+<mat-card class="preference-group">
+  <label class="header" [class.enabled]="enabled" [class.always-enabled]="alwaysEnabled">
   <span class="title">
     {{name}}
   </span>
-  <app-switch [(value)]="value" *ngIf="!alwaysEnabled"></app-switch>
```
jcc37, line 5: You may remove `SwitchModule` and its files since it is no longer used by any other components.

64json: oOops, I forgot to include the removal in git commit. Thanks for pointing that out.

PR #45, opened 2020-06-29
Render multiple legend items on line chart

64json:

PR #46, opened 2020-06-30
Multiple datasets support

jcc37:
Branched off of #44 

- Add `dataset` types and interfaces to expose general APIs to retrieve data from different sources
- Refactor `PreferenceSerice` and support "select" type preference item
- Modify `DataService` and `Preference` to support switching between different sets of dataset preference
- Merge #26 to #30 and update old tests for the new code structure and data flow
- Add new tests for component and dataset

What's next
- Refactor `UserWhiteNoiseDataset`
  - Create util functions to create dataset query function with lazy-query and caches
  - Move common category generator to utils, like `nthDay` in `DataCube`
- Refactor `XYChartD3`
  - Support data points with different `x` and `y` types for other chart components


src/datasets/types.ts
```diff
@@ -0,0 +1,50 @@
...
+  y: U;
+}
+
+export interface ChartMeta<Type, DataGeneratorType> {
+  type: Type;
+  fetchs: DataGeneratorType;
+  title: string;
+  xlabel?: string;
+  ylabel?: string;
+}
```
64json, line 23: I think naming them `ChartMetaType` and `ChartMeta` is quite confusing since `Type` suffix is generally used to indicate a constructor type.

In addition, I think it's better to remove `Meta` suffix from both. It makes more sense for `Dataset.charts` to be of type `Chart[]` than `ChartMetaType[]`.

So my suggestion is to rename `ChartMetaType` to `Chart`, and `ChartMeta` to `BaseChart`.

src/datasets/types.ts
```diff
@@ -0,0 +1,50 @@
...
+export type LineChartMeta = ChartMeta<'line', LineChartDataGenerator>;
+
+export type LineChartDataGenerator = (options: LineChartQueryOptions) => LineChartData[];
+
+export interface LineChartQueryOptions {
+  range: [Date, Date];
+}
+
+export interface LineChartData {
+  data: Datum<Date, number>[];
```
64json, line 40: `data` property in `Data` sounds confusing. How does renaming `LineChartData` to `LineChartLegendItem` sound?

src/datasets/types.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export interface TabbedChartsMeta {
+  type: 'tabbed';
+  charts: ChartMetaType[];
+  title: string;
+}
+
+export type LineChartMeta = ChartMeta<'line', LineChartDataGenerator>;
+
+export type LineChartDataGenerator = (options: LineChartQueryOptions) => LineChartData[];
```
64json, line 33: I think it should be named `LineChartQuery` to be consistent with `LineChartQueryOptions`.

src/d3/xy-chart.d3.ts
```diff
@@ -2,15 +2,14 @@ import * as d3 from 'd3';
...
 import { Observable } from 'rxjs';
 import { formatX } from '../utils/formatters';
+import { XYChartData, Datum } from '../datasets/types';
 
-export interface Datum {
-  date: Date;
-  value: number;
-}
+// TODO: Generic Datum type
+export type DatumDN = Datum<Date, number>;
```
64json, line 8: nit: probably not use abbreviation unless very common?

I can tell DN stands for Date and Number from its definition, but it would be hard to tell if I didn't know the definition. Also, because DatumDN inherits Datum, it makes more sense to keep `Datum` as suffix than prefix.

src/datasets/types.ts
```diff
@@ -0,0 +1,50 @@
...
+export type ChartMetaType = LineChartMeta | TabbedChartsMeta;
+
+export type XYChartData = LineChartData;
+
+export interface Datum<T, U> {
+  x: T;
+  y: U;
+}
+
+export interface ChartMeta<Type, DataGeneratorType> {
```
64json, line 17: `Type` should extend `string`.

src/datasets/types.ts
```diff
@@ -0,0 +1,50 @@
...
+  type: Type;
+  fetchs: DataGeneratorType;
```
64json, line 19: How about renaming it to `query: Query`?

src/datasets/user-wgn.dataset.ts
```diff
@@ -0,0 +1,176 @@
...
+  avgHits: 10000,
+  hitStdDev: 100,
+  avgUsers: 100,
+  userStdDev: 1,
+  avgSessionsPerUser: 5,
+  sessionsPerUserStdDev: 3,
+};
+
+// TODO: Refactoring
+export function UserWGNDataset(config: UserWGNDatasetConfig): Dataset {
```
64json, line 46: What does WGN stand for?

src/datasets/types.ts
jcc37: How about renaming `Datum` to `XYPoint` and `LineChartData.data` to `LineChartData.points`?

src/datasets/types.ts
jcc37: Good idea! I will do that.

src/d3/xy-chart.d3.ts
jcc37: I would like to let `XYChartD3` and `RenderOptions` take type variables for x and y, since it will later be extended by other charts like `BarChart`. However, we may also need to take different format functions (value to string) for different types. Do you have any suggestions for it?

src/datasets/user-wgn.dataset.ts
jcc37: White Gaussian Noise. I admit that it is a bad name, and we should rename it. Any good ideas?

src/datasets/types.ts
64json: Yeah that sounds good too 👍

src/d3/xy-chart.d3.ts
64json: Hmm good point, should we have them as properties of `XYChartMeta`?

```ts
interface XYChartMeta<X, Y> extends ChartMeta<...> {
  formatX(value: X): string;
  formatY(value: Y): string;
}
```

64json: I think y will always be numeric while x should be generic.

src/datasets/user-wgn.dataset.ts
64json: `UserWhiteGaussianNoiseDataset` 😬 sounds pretty long, but I think it's okay since we would type its name in not many places.

Speaking of it, how do you think of creating a namespace (or class with static properties) to omit the long prefix?

```ts
namespace UserWhiteGaussianNoiseDataset {
  export interface Config { ... };

  export const defaultConfig: Config = { ... };

  export function create(): Dataset { ... };
}
```

src/d3/xy-chart.d3.ts
jcc37: I think letting `XYChartMeta`  bring the formatters makes sense.

src/datasets/user-wgn.dataset.ts
jcc37: I can rename the dataset to `UserWhiteNoiseDataset` to make it shorter. What I mean is the name does not well describe the feature pattern of this dataset.

I think using `namespace` is not a good pattern in TypeScript (I remember there are some posts talking about it?), but I don't believe class with static methods is suitable for this case either. Another way is to take advantage of the module:
```typescript
// in dataset.ts
export interface config { ... };
export const defaultConfig { ... };
export function create(): Dataset { ... };

// in other ts files
import * as UserWhiteNoiseDataset from '...'
```
but it will be painful to specify the name of dataset in every ts file.

64json: I don't think we will directly type the dataset name in many places other than in `DataService` and `DashboardComponent`, right? I think the usage of it as in your snippet looks good enough.

src/datasets/types.ts
```diff
@@ -0,0 +1,58 @@
...
+export interface ComponentMeta<T extends string> {
+  type: T;
+  title: string;
+}
+
+export interface DataComponentMeta<T extends string, QueryT> extends ComponentMeta<T> {
+  query: QueryT;
+}
+
+export interface ContainerComponentMeta<T extends string, MetaT> extends ComponentMeta<T> {
```
64json, line 19: `MetaT extends (Meta | Meta[])`?

I think it would be cleaner to have two separate interfaces for singular and plural metas respectively.

jcc37: I assume `ContainerComponentMeta` can take any object composed of structure information and `Meta`s. For example, a `GridChartsMeta` may take a 2-dimensional array of `Meta`.

However, I don't know if there will be other container components in the future. It may be overengineered to take it into consideration.

src/datasets/types.ts
jcc37: I have refactored the types based on your reviews and our discussion. Feel free to comment on the new commits.

src/datasets/user-wgn.dataset.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  const makeLineChartMeta = (measureName: string): LineChartMeta => {
+    const rows = data.map(row => ({
+      x: row.categories.date,
+      y: row.values[measureName],
+    }));
+
+    const query = (options: LineChartQueryOptions) => ([{
+      points: rows.filter(xBetweenDates(...options.range)),
+    }]);
```
64json: I think the data cube querying behavior (line 59 - 107) should be happening within this `query` method.

When we think of `DataCube` as the backend server of GA, it's like fetching the entire data without knowing which portion of the data is actually needed.

src/datasets/types.ts
64json: I was mostly concerned of naming. A container of a single meta should have `meta`, that of multiple metas should have `metas`, that of `Map<Meta>` should have `metaMap`, etc.

With that being said, is there a reason that we need `ContainerComponentMeta`? I think it's okay to not have a common parent interface unless we need to group them for their common behavior.

src/datasets/user-wgn.dataset.ts
jcc37: It is for performance concerns. As `queryOptions` has no effect on `rows` (line 103), we can prepare them before the `query` function is called and reuse them when applying different `queryOptions`. We just need to "mock" the query service.

By the way you made me notice that I should move line 103 outside function `createLineChartMeta`.

src/datasets/types.ts
jcc37: Alright, I agree there should not exist a `ContainerComponentMeta` as parent. I will remove it, thanks!

src/datasets/user-wgn.dataset.ts
64json: Let's say the range options are 1 month and 1 year long. Are we going to query in advance for the entire year data when the initial range option is set to be only 1 month long?

64json: GA similarly makes a new API call (or "query") to the server (or "data cube") every time the range changes.

64json: Considering users don't change range options as often, initially querying small amount of data would rather enhance performance.

jcc37: For this dataset: yes. Since these implementations have similar memory consumption and are both correct, which means the `query` function produces exact same output, I would choose the one with less execution time. Moreover, other datasets can have their own `query` implementation.

jcc37: Or we can have some utilities to create query function with lazy query and caches.

64json: Both do behave correctly, but the current implementation would consume way more memory.

To roughly estimate, the current implementation would have `rawData` with the length of 365, and my suggested implementation would have the length of 30.

Even if they were to have similar memory consumption, I would mimic how GA is handling it, considering the goal of this project.

64json: > Or we can have some utilities to create query function with lazy query and caches later. What do you think?

Yeah I agree that we should make it a util since many of datasets/metas would have a similar query function.

jcc37: > Both do behave correctly, but the current implementation would consume way more memory.
> 
> To roughly estimate, the current implementation would have `rawData` with the length of 365, and my suggested implementation would have the length of 30.
> 
> Even if they were to have similar memory consumption, I would mimic how GA is handling it, considering the goal of this project.

The memory consumption should be dominated by object `dataCube`. That's why I said similar memory consumption.
 
The real-world GA data can be used in this project by creating a dataset that sends requests to the backend service in query functions. It has nothing to do with the fake dataset generated by data generator.

64json: It is true that the memory consumption is dominated by `dataCube`, but by querying the entire `dataCube`, the current implementation doubles the memory consumption.

jcc37: > > Or we can have some utilities to create query function with lazy query and caches later. What do you think?
> 
> Yeah I agree that we should make it a util since many of datasets/metas would have a similar query function.

How about working on that in a separate PR? After you merge it to #45 and add those compounded measures.
You can modify the implementation when merging it, and I also hope you to refactor the `create` function consider the change of the entire design.

64json: Okay, I'll refactor the `create` function in #45. Are you planning to make further updates to `user-white-noise.dataset.ts` in this PR?

jcc37:
No. I will work on modifying PreferenceService, DataService, and UI
components to select datasets. I would also make the changes of formatX,
formatY  in a separate PR.

src/datasets/user-wgn.dataset.ts
64json: Actually I think it's better to refactor the function in a separate branch (#X), since this refactoring would be far from what #45 is focusing on. I'll merge #X into this PR under your review, and then merge this PR into #45.  How do you think?

jcc37: Great. I have added cards to the project board.

src/d3/xy-chart.d3.ts
jcc37: This feature will be added in a separate PR.

src/components/card/card.component.spec.ts
```diff
@@ -1,28 +1,84 @@
...
+    lineComponent.meta = testChartMeta1;
+    lineComponent.ngOnInit();
+    expect(lineComponent.currentChart.title).toBe(testChartMeta1.title);
+
+    tabbedComponent.meta = testTabbedChartsMeta;
+    tabbedComponent.ngOnInit();
+    expect(tabbedComponent.currentChart.title).toBe(testChartMeta1.title);
+  });
+
+  it('should return whether input meta is tabbed charts or not.', () => {
```
Rhombus33, line 59: Would be better to test the user facing behavior instead of part of the implementation (i.e. if the tabs div is present in the DOM)

src/components/card/card.component.spec.ts
```diff
@@ -1,28 +1,84 @@
...
+    lineComponent.meta = testChartMeta1;
+    expect(lineComponent.isTabbed()).toBeFalse();
+    tabbedComponent.meta = testTabbedChartsMeta;
+    expect(tabbedComponent.isTabbed()).toBeTrue();
+  });
+
+  it('should return input meta titles.', () => {
```
Rhombus33, line 66: Same here and throughout this test file. Better to test the feature and if it appears correctly in the DOM rather than the implementation. For this case test to make sure the DOM is updated with the correct titles.

src/components/card/card.component.ts
```diff
@@ -1,26 +1,48 @@
...
+    }
+  }
+
+  setCurrentTabTitle(title) {
+    this.currentTabTitle = title;
   }
 
-  setMeasureName(measureName) {
-    this.currentMeasureName = measureName;
+  isTabbed(): this is CardComponent<TabbedChartsMeta> {
```
Rhombus33, line 45: Can this type just be boolean?

src/components/card/card.component.ts
```diff
@@ -1,26 +1,48 @@
...
-  @Input() tabbed = false;
-  @Input() type: CardType;
-  @Input() measureNames: string[];
+export class CardComponent<T extends Meta> implements OnInit {
+  @Input() meta: T;
   humanizeMeasureName = humanizeMeasureName;
-  currentMeasureName: string;
+  currentTabTitle: string;
+
+  get titles() {
```
Rhombus33, line 16: add type?

src/components/line-chart-audification/line-chart-audification.component.spec.ts
```diff
@@ -21,26 +22,31 @@ describe('LineChartAudificationComponent', () => {
...
       ],
     });
     const hostFixture = TestBed.createComponent(LineChartComponent);
     const host = hostFixture.componentInstance;
-    host.measureName = 'activeUsers';
-    host.ngOnChanges({ measureName: new SimpleChange(null, host.measureName, true) });
+    const time = new Date();
+    host.meta = createLineChartMeta(
+      'Title',
+      () => [{
```
Rhombus33, line 36: Could we define this callback separately and reference it here? It will make it easier to know what it does/ is for.

src/components/card/card.component.ts
64json: If it were `boolean` type instead of the type guard, inside the if-statements (line 18, 26, 35) would need unnecessary type assertions like `return (this.meta as TabbedChartsMeta).metas.map(c => c.title);`.

src/components/line-chart-audification/line-chart-audification.component.spec.ts
```diff
@@ -21,26 +22,31 @@ describe('LineChartAudificationComponent', () => {
...
+        label: '',
+        points: [{ x: time, y: 0 }, { x: new Date(time.getTime() + 10), y: 1 }],
+      }],
+    );
 
     TestBed.resetTestingModule();
     TestBed.configureTestingModule({
       imports: [
```
jnager, line 44: Apologies if this is just me not understanding TestBed (we use a slightly different wrapper around TestBed), but most of our tests import the module under test, which would include the dependencies like ScreenReaderModule or MatCardModule. So our setup for this would be 
```
imports: [
  LineChartAudificationModule,
],
providers: [...]
```
And we would not declare the component. Is this not the recommended pattern for TestBed setup?

src/components/line-chart/line-chart.component.ts
```diff
@@ -69,21 +75,13 @@ export class LineChartComponent implements RenderOptions, OnChanges, OnInit, OnD
...
-          return [];
-        }
-        const [firstMeasureName] = Object.keys(rows[0].values);
-        return rows.map(row => ({
-          date: row.categories.date,
-          value: row.values[firstMeasureName],
-        }));
+      .pipe(map(queryOption => {
+        this.activePoint$.next(null);
+        return this.meta.query(queryOption)[0];
```
jnager, line 82: Could you add a comment explaining the significance of index 0 here?

src/services/preference/preference.service.ts
```diff
@@ -40,22 +79,34 @@ export class PreferenceService {
...
+      if (!(key in preferenceMeta)) {
         delete loadedPreference[key];
         continue;
       }
 
-      const expectedType = typeof defaultPreference[key];
+      const expectedMeta = preferenceMeta[key];
       const actualType = typeof loadedPreference[key];
 
       // in case of type mismatch, override with the default value
```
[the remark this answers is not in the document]
jcc37: What do you mean? I think this comment is relevant to the code below.

src/components/dashboard/dashboard.component.spec.ts
```diff
@@ -1,19 +1,30 @@
...
 
 describe('DashboardComponent', () => {
   let fixture: ComponentFixture<DashboardComponent>;
   let component: DashboardComponent;
+  let preferenceService: PreferenceService;
+  let dataService: DataService;
 
   beforeEach(() => {
+    preferenceService = new PreferenceService();
+    dataService = new DataService(preferenceService);
```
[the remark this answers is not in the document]
jcc37: I think we can assume these services work as expected when testing components.

src/components/line-chart/line-chart.component.ts
jcc37: Sure, I will add a comment to it.

It is just a workaround. `LineChartComponent` should render multiple lines when there are more than one datum in the array, but it is not able to do that now. This feature will be supported in #45 .

src/components/card/card.component.spec.ts
jcc37: Then do we need to keep these unit tests on return values?

src/components/card/card.component.spec.ts
```diff
@@ -1,28 +1,84 @@
...
+  const testChartMeta2: LineChartMeta = {
+    type: 'line',
+    title: 'testChart2',
+    query: () => [],
+  };
+  const testTabbedChartsMeta: TabbedChartsMeta = {
+    type: 'tabbed',
+    title: 'testTabbedCharts',
+    metas: [testChartMeta1, testChartMeta2],
+  };
```
64json, line 27: How do you think of moving these to `src/utils/mocks.spec.ts`?

src/components/card/card.component.ts
Rhombus33: Oh I see, I did not make that connection. Cool but hidden type guard!

Could this be a bit cleaner if the CardComponent could take in an input of chartMeta arrays or something like a CardMeta instead of any Meta. The CardMeta could just have an array of ChartMetas and any other card level information. This way you don't need to narrow from Meta to TabbedChartsMeta. And to avoid the split logic for many of the get methods, could we just keep a reference to the current chart meta?

lmk what you think

src/components/card/card.component.spec.ts
Rhombus33: I think these are small enough that they should stay here. Makes it easier to reference them when reading the tests.

I would only move test data if it is HUGE or if it is shared by more than one test file

src/components/line-chart-audification/line-chart-audification.component.spec.ts
64json: That sounds better, so we don't need to re-specify the dependencies for testing. I'll work on it to see if it works.

jcc37: I think both are reasonable. These tests started from the default `spec.ts` generated by Angular CLI, and it declares the components in TestBed. Besides, the these test files are named `xxx.component.spec.ts`, which stand for tests for `xxx.component.ts`, so it is reasonable to declare the component directly.
I think a better way is to move the test to module level, like renaming it to `xxx.module.spec.test`

src/components/card/card.component.ts
jcc37: Sounds good, does it mean `ChartMeta` will always be wrapped in some kinds of `CardMeta`?
When designing this part, I assume that we can have nested `CardMeta`, like a grid card with one of block being a tabbed card (charts). In this case, we still need to have some logic inside card component to narrow from `Meta` to `ChartMeta` or CardMeta. A better practice is to make these card into different CardComponents, like `TabbedCardComponent` and `SingleChartCardComponent`, but it will be lots of refactoring works and a separate PR.
I dont know if my assumption and implementation is over engineering. Maybe we just need some very simple card structure.

src/components/card/card.component.spec.ts
64json: ok sounds good. I'll move it once I find another place to use it.

src/components/line-chart-audification/line-chart-audification.component.ts
```diff
@@ -101,7 +102,7 @@ export class LineChartAudificationComponent implements AudificationPreference, O
     } else if (key === 'l') {
       this.readOutMeasure();
     } else if ('0' <= key && key <= '9') {
-      const datumIndex = Math.floor(+key / 10 * this.data.length);
+      const datumIndex = Math.floor(+key / 10 * this.points.length);
```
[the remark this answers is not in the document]
64json: I couldn't think of a simple name for the `10`, so I broke `+key / 10` down into `seekPercentage / 100` where `seekPercentage = (+key * 10)`.

src/components/preference-item/preference-item.component.spec.ts
```diff
@@ -1,37 +1,112 @@
...
+      imports: [
+        MatCardModule,
+      ],
       declarations: [
         PreferenceItemComponent,
       ],
     });
-    fixture = TestBed.createComponent(PreferenceItemComponent);
-    component = fixture.componentInstance;
+    boolFixture = TestBed.createComponent<PreferenceItemComponent<TestPreference, 'bool'>>(PreferenceItemComponent);
```
jnager, line 31: Same comment here about simplifying these tests into distinct describe blocks (which can be done in a later PR if preferred).

src/components/preference-item/preference-item.component.spec.ts
```diff
@@ -1,37 +1,112 @@
...
+        bool: {
+          type: 'boolean',
+          defaultValue: true,
+        },
+        select: {
+          type: 'select',
+          defaultValue: 'option1',
+          options: ['option1', 'option2'],
+        }
+      },
```
jnager, line 61: Just to reiterate my comment from the Preference types doc below, I think Preferences would be easier to read without the `_meta` being stored in a distinct object separate from the current value.

src/components/preference-item/preference-item.component.html
```diff
@@ -5,5 +5,10 @@
   <ng-container [ngSwitch]="type">
     <input *ngSwitchCase="'number'" type="number" [(ngModel)]="value"/>
     <mat-slide-toggle *ngSwitchCase="'boolean'" [(ngModel)]="value"></mat-slide-toggle>
+    <mat-select *ngSwitchCase="'select'" [(ngModel)]="value">
+      <mat-option *ngFor="let option of options" [value]="option">
+        {{option}}
+      </mat-option>
+    </mat-select>
```
jnager, line 12: Just for future PRs - Adding the mat-select and modifying the PreferenceGroup / PreferenceItem would be a nice distinct PR for review. Totally get that sometimes features come up in the middle of other tasks, but just wanted to note this for the future (mostly so we can hopefully churn through reviews a bit faster :D ). Thanks!

src/datasets/queries/time-series.query.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export type LegendItem<S> = {
+  label: string;
+  style?: Partial<S>;
+  measureName: string;
+  periodOffset?: number;
+  windowSize?: number;
+};
+
+export function createTimeSeriesQuery<S>(dataCube: DataCube, legendItems: LegendItem<S>[]): TimeSeriesQuery<S> {
```
jnager, line 30: Please add a JSDoc for this exported function.

src/services/data/data.service.spec.ts
```diff
@@ -1,27 +1,16 @@
 import { DataService } from './data.service';
-import { mockMeasureName } from '../../utils/mocks.spec';
-import { Duration } from 'luxon';
+import { PreferenceService } from '../preference/preference.service';
 
 describe('DataService', () => {
```
jnager, line 4: Could you please add tests for the DataService functionality (i.e. mock the preferenceService providing dataset$)? Can be done in later PR if needed.

src/services/preference/types.ts
```diff
@@ -1,7 +1,12 @@
+import * as UserWhiteNoiseDataset from '../../datasets/user-white-noise.dataset';
+import * as DummyDataset from '../../datasets/dummy.dataset';
+
 export interface Preference {
   enabled: boolean;
 }
 
+export type PreferenceWithMeta<T extends Preference> = T & { _meta: PreferenceMeta<T> };
```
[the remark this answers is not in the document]
jcc37: Since almost all of the components only take and update current preference values and have nothing to do with the meta-information (default value, type, and other things like options for select type), separating them would be more convenient for us when developing components and directives to access `PreferenceService`.

For example, we can update the preference value without worrying about the existence of `_meta`:
https://github.com/googleinterns/guide-doge/blob/99a4025fe31cb404600e7d6c059043aab656f324/src/components/preference-group/preference-group.component.ts#L24

src/components/preference-item/preference-item.component.html
jcc37: No problem, sorry for the inconvenience. I will break them into smaller PRs in my future work.

src/datasets/queries/time-series.query.ts
jcc37: @jnager Sure. A related question: what kinds of exported functions need JSDoc? Is it necessary to add JSDoc to all the exported functions?

@64json would you mind me to do that?

64json: @jcc37 I'd appreciate if you can work on it!

src/components/card/card.component.spec.ts
jcc37: Moved to #54 . Will be resolved in separate PRs.

src/components/line-chart-audification/line-chart-audification.component.spec.ts
jcc37: Moved to #54 . Will be resolved in separate PRs.

src/components/card/card.component.spec.ts
```diff
@@ -1,28 +1,84 @@
...
       declarations: [
         CardComponent,
       ],
+      imports: [
+        MatCardModule
+      ]
     });
-    fixture = TestBed.createComponent(CardComponent);
-    component = fixture.componentInstance;
+    tabbedFixture = TestBed.createComponent<CardComponent<TabbedChartsMeta>>(CardComponent);
```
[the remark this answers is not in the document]
jcc37: Moved to #54 . Will be resolved in separate PRs.

src/components/line-chart-audification/line-chart-audification.component.spec.ts
jcc37: Moved to #54 . Will be resolved in separate PRs.

src/services/data/data.service.spec.ts
jcc37: Oh, I should have done that. I will add these tests in this PR in the old way, and then refactor them when resolving #54 .

src/datasets/queries/time-series.query.ts
jnager: Our internal styleguide requires JSDoc comments for every top-level export of a module and generally recommends comments for any method or properties whose purpose is not obvious from name/type (often judged by reviewers).

I don't think you need to add JSDoc comments on every function since this is external to our codebase, but I personally find comments more helpful than not usually, and I do think they helps reviews go faster!

src/components/dashboard/dashboard.component.spec.ts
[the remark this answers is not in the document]
jnager: In general we try to mock almost all dependencies in order to not introduce potentially-breaking dependencies into unit tests, which ideally test only the single component (i.e. Dashboard) and do not need to care about changes made to PreferenceService or DataService. It also helps make the test a better form of documentation - i.e. I can clearly see from a test what I expect PreferenceService or DataService to provide to the Dashboard.

Perhaps this is different however - I have not looked too carefully at the dashboard code to see how / if these services are used. Feel free to ignore if I'm missing stuff.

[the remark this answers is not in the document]
64json: I agree it would be ideal. Do the mocked services extend the actual services and override methods to make them do nothing? Also, if we end up mocking every dependency, should we still import the module (as we discussed [here](https://github.com/googleinterns/guide-doge/pull/46#discussion_r452401076))?

src/services/preference/types.ts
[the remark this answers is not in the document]
jnager: Gotcha - sounds good.

src/components/dashboard/dashboard.component.spec.ts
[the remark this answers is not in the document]
jnager: Yep, I think you would still import the module and then provide the mocked services in the `providers` array like you already are but the "service" would be something like:
```
let preferenceService: jasmine.SpyObj<PreferenceService>;

beforeEach(() => { 
  preferenceService = new jasmine.createSpyObj('preferenceService', ['data$, 'audification$', ...etc.]);

 ...

  providers: [
    { provide: PreferenceService, useValue: preferenceService,}
  ]
  ...
```

It is a bit of extra code and kind of annoying to set up, but does allow for more isolated unit tests.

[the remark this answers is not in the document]
64json: Ahh, didn't know you can create an entire spy object like that. That would be very useful. We will address it in #54, thanks!

src/components/card/card.component.ts
Rhombus33: I think it's okay to leave it as is! Perhaps add a comment explaining that this is a type guard as well!

src/services/preference/preference.service.ts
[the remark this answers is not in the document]
Rhombus33: I believe the comment above refers to line 98 now

src/services/data/data.service.spec.ts
jcc37: Part of the tests has been added, while others may need some mock datasets or refactoring. The remaining tests should be added when resolving #54

src/datasets/metas/types.ts
```diff
@@ -0,0 +1,23 @@
+import { LineChartMeta } from './line-chart.meta';
+import { TabbedChartsMeta } from './tabbed-charts.meta';
+
+export interface BaseMeta<T extends string> {
+  type: T;
+  title: string;
+}
+
+export interface DataMeta<T extends string, QueryT> extends BaseMeta<T> {
+  query: QueryT;
```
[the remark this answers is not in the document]
jcc37: I prefer to keep the verb "query" here, but naming it "queryData" sounds good.

src/services/preference/preference.service.ts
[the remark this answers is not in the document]
jcc37: It refers to the code between line 91 and 100. The "select" type here can be written as `"option1" | "option2" | ... ` in TypeScript, but there is no easy solution to do this type check in runtime. We need to verify the type to be a string and be one of the options so that we can determine if it is a type mismatch.

src/datasets/user-white-noise.dataset.ts
```diff
@@ -0,0 +1,95 @@
...
+    type: 'number',
+    defaultValue: 5,
+  },
+  sessionsPerUserStdDev: {
+    type: 'number',
+    defaultValue: 3,
+  },
+};
+
+export function create(config: Config): Dataset {
```
Rhombus33, line 54: You don't need to do this in this PR, but generally it's nice to have a small documentation for exported methods, services or any significant chunk of code so that someone can get the gist of it without reading through all the code

src/components/card/card.component.ts
Rhombus33: Something like "Type guard for narrowing down meta type to "TabbedChartsMeta""

src/datasets/user-white-noise.dataset.ts
jcc37: It raises another design question here. Actually `Config`, `configMeta`, and `create` are three necessary exports/properties for all dataset module/object, it is nice to have a abstract parent interface or class for all datasets to extend. However, they would be all static and immutable properties and I don't want developers to create instance of dataset if I make it a class, so I make them modules when adding dataset support.
The bad things of this design are that it is hard to build mock datasets for testing and do the type inference on the map of dataset modules:
https://github.com/googleinterns/guide-doge/blob/0b203248399232462673de30cdcddd2e72608dd9/src/datasets/index.ts#L4

I would like to refactor it if you have any good ideas.

64json: Off-topic but it would be great if TypeScript supports module extending interface.

Something like `declare module extends AbstractDataset;` and all the members of `AbstractDataset` has to be implemented and exported, so people don't need to have an extra wrapper to make it type-safe.

Issue #41, opened 2020-06-25
Make LineChart to take multiple legend items


64json:

64json:
Closed by #45

PR #48, opened 2020-07-06
Refactor query and meta

64json:
1. Make queries reusable by different metas
    - e.g., `LineChartMeta` and `BarChartMeta` can both use `TimeSeriesQuery`.

2. Move types to appropriate files and rename accordingly


src/d3/xy-chart.d3.ts
```diff
@@ -2,14 +2,12 @@ import * as d3 from 'd3';
...
-
-// TODO: Generic Datum type
-export type DNPoint = XYPoint<Date, number>;
+import { TimeSeriesPoint } from '../datasets/queries/time-series.query';
+import { LineChartDatum } from '../components/line-chart/line-chart.component';
 
 export interface RenderOptions extends BaseRenderOptions {
-  data$: Observable<XYChartData>;
-  activeDatum$: Observable<DNPoint | null>;
+  data$: Observable<LineChartDatum>;
```
jcc37, line 9: My general advise for naming variables in this project:

`point` for [x, y] pair data
`points` for list of point (because data points)
`data` for object of points and other related information (and use `data.points` to get list of point to plot)
and please align variable names with their types. A variable with `data` in its name should be in `Data` type.

Although "datum" is the singular form of "data", the later is used more often and are usually even used in both cases. I would suggest to use "data" whenever possible in this project.

src/datasets/queries/time-series.query.ts
```diff
@@ -0,0 +1,110 @@
...
+  label: string;
+  style: Partial<S>;
+  points: TimeSeriesPoint[];
+}
+
+export type TimeSeriesQuery<S> = (options: TimeSeriesQueryOptions) => TimeSeriesDatum<S>[];
+
+export type LegendItem<S> = {
+  label: string;
+  style: Partial<S>;
```
jcc37, line 24: Why not making `style` an optional property?

src/datasets/user-white-noise.dataset.ts
```diff
@@ -41,135 +31,47 @@ export const defaultConfig: Config = {
...
+    measures.map(measure => {
+      const label = humanizeMeasureName(measure.name);
+      return {
+        type: 'line',
+        title: label,
+        query: createTimeSeriesQuery(dataCube, [{
+          label,
+          measureName: measure.name,
+          style: {},
+        }]),
```
jcc37, line 54: use `createLineChartMeta`

src/utils/misc.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Returns an array with unique values
+ *
+ * @param array The input array with possibly redundant values
+ */
+export function unique<T>(array: T[]): T[] {
+  return [...new Set(array)];
+}
```
jcc37, line 8: A general question: why don't we use `lodash` and `moment.js` in our project?

src/d3/xy-chart.d3.ts
64json: I agree on "point" and "points". It is true that `data` is much more commonly used as a collective noun, but I would still lean on calling it "datum" so that we can address `Datum[]` as `Data`. (i.e., naming it `datas` is grammatically wrong.)

I'll align the variable names with their types in #45 also, to prevent massive merge conflicts.

src/datasets/user-white-noise.dataset.ts
64json: good point

src/utils/misc.ts
64json: I read multiple negative articles about using `lodash` including https://codeburst.io/why-you-shouldnt-use-lodash-anymore-and-use-pure-javascript-instead-c397df51a66.

I've only needed simple addition/subtraction to Date objects, and d3 handled the formatting part. I don't have much objection to using date/time libraries to improve readability for more complicated manipulation or formatting.

Also, you might be interested in using [`luxon`](https://moment.github.io/luxon/), which is derived from `moment.js`.

src/datasets/user-white-noise.dataset.ts
```diff
@@ -9,19 +7,11 @@ import {
...
-import {
-  CompoundMeasure,
-  CompoundMeasureType,
-  destructureCompoundMeasure,
-  isCompoundMeasure,
-  namePeriodOverPeriodMeasure,
-  nameRollingMeasure,
-  PeriodOverPeriodMeasure,
-  RollingMeasure,
-} from '../utils/compoundMeasures';
```
jcc37, line 24: You can now remove `utils/compoundMeasures.ts`

src/components/line-chart/line-chart.component.ts
```diff
@@ -27,13 +30,15 @@ export class LineChartComponent implements RenderOptions, OnChanges, OnInit, OnD
...
-  queryOptions$ = new BehaviorSubject<LineChartQueryOptions>({
+  queryOptions$ = new BehaviorSubject<TimeSeriesQueryOptions>({
     range: [this.startDate, this.endDate],
   });
-  data$ = new BehaviorSubject<LineChartData>({
+  data$ = new BehaviorSubject<LineChartDatum>({
+    label: '',
     points: [],
+    style: {},
   });
```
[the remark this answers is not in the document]
64json: I'll address this in #41

PR #51, opened 2020-07-07
Fix base path of gh-pages deployment

64json:


64json:
Pushed the same commit to #37, hence closing.

PR #53, opened 2020-07-08
Audification multiple legend items support

64json:


src/components/card/card.component.scss
```diff
@@ -4,6 +4,7 @@
   display: flex;
   flex-direction: column;
   align-items: stretch;
+  width: 854px;
```
Awk34, line 7: magic number?

src/components/line-chart-audification/line-chart-audification.component.spec.ts
```diff
@@ -106,6 +102,14 @@ describe('LineChartAudificationComponent', () => {
     }
   });
 
+  it('should switch the legend item upon pressing UP or DOWN.', () => {
+    ['ArrowUp', 'ArrowDown'].forEach(key => {
+      const prevDatumIndex = component.datumIndex;
+      triggerKeyDown(key);
+      expect(component.datumIndex).not.toBe(prevDatumIndex);
```
Awk34, line 109: should you also be checking that the legend item changes focus?

src/i18n/utils.ts
```diff
@@ -32,5 +32,10 @@ export function t(key: I18nKey, ...args: any[]) {
  * @param args The arguments to format the string with.
  */
 export function tA11y(key: I18nKey, ...args: any[]) {
-  return striptags(t(key, ...args));
+  return striptags(
+    t(key, ...args)
+      .replace(/\.{3}/g, 'to')
+      .replace(/<kbd>(\S+)<\/kbd>/g, `'$1'`)
+      .replace(/'\?'/g, `'QUESTION MARK'`),
```
Awk34, line 39: Is it a problem that these are always english?

src/components/card/card.component.scss
64json: I'll make a variable for the magic number.

src/i18n/utils.ts
64json: oh I think it should be in i18n indeed.

src/components/line-chart-audification/line-chart-audification.component.spec.ts
64json: Which element do you mean by the legend item?

Awk34: The test describes checking that UP/DOWN changes the legend item, but it looks like the test just verifies that new index is not the old index. Should it maybe verify that the new index is the correct to focus?

PR #56, opened 2020-07-11
Web vr add data2

cvillapudua:
- added `dataService` and `preferenceService` to `VRScatterPlotComponent`
- both services allow to connect to data through `TimeSeriesQuery`
- fix broken tests (need to refactor tests to parallel new data format taken by `scatterplot.d3.ts`


src/components/vr-scatter-plot/vr-scatter-plot.component.ts
```diff
@@ -1,32 +1,95 @@
...
+    dataService.dataset$
+    .pipe(takeUntil(this.destroy$))
+    .subscribe(dataset => {
+      // componentMetas is initialized to different dataset metas - will help funnel dataset
+      this.componentMetas = dataset.metas;
+      // dataset.metas[0].type = 'tabbed' and dataset.metas[1] = 'line'
+      this.meta = dataset.metas[0];
+      if (this.isTabbed()) {
+        this.meta2 = (this.meta as any).metas[0];
+        console.log(this.meta2);
```
mfitz3, line 51: Can remove.

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
```diff
@@ -1,32 +1,95 @@
...
+
+  isTabbed(): boolean {
+    return 'tabbed' === this.meta.type;
+  }
+
+  get datum() {
+    return this.datum$.value;
+  }
+
+  sleep(milliseconds: number) {
```
mfitz3, line 86: Why is this needed? Can you use `setTimeout` instead?

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
```diff
@@ -1,32 +1,95 @@
...
+import { BehaviorSubject, Subject } from 'rxjs';
+import { DAY } from '../../utils/timeUnits';
+import { map, takeUntil } from 'rxjs/operators';
 
 @Component({
   selector: 'app-vr-scatter-plot',
   templateUrl: './vr-scatter-plot.component.html'
 })
-export class VRScatterPlotComponent implements OnInit, OnChanges, OnDestroy{
+export class VRScatterPlotComponent<T extends Meta>implements OnInit, OnChanges, OnDestroy{
```
mfitz3, line 17: Add test for this component.

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
cvillapudua: @mfitz3 do you just mean the `console` printing statement?

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
cvillapudua: I used this during debugging because I there was a timing/sync problem -- I'm not super sure, but from what I read on component life cycles in Angular the `constructor` and `onInit` functions could be running at the same time... however for my design I need everything in the constructor to be declared/initialized before `onInit` is called.

My solution was to put the meta handling of `DataService` in `onInit` and then at the end of the function, I called the second chunk of code, where I access `queryOptions` of the meta selected in `onInit`. This worked, but please let me know if you have any suggestions!

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
cvillapudua: just added testing in the most in https://github.com/googleinterns/guide-doge/pull/56/commits/63d9a7efca75808c621de8cb82b664dfe44b0fe6

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
mfitz3: Yes

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
mfitz3: `constructor` should be called prior to `ngOnInit `. 

Two thoughts here that might help:
1. You could wrap the contents of `init` in a `setTimeout` call with a small delay.
2. You could try using another lifecycle hook (e.g. `ngAfterViewInit`) that runs later, instead of `ngOnInit`.  See: https://angular.io/guide/lifecycle-hooks

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
```diff
@@ -1,32 +1,88 @@
...
+      // componentMetas is initialized to different dataset metas - will help funnel dataset
+      this.componentMetas = dataset.metas;
+      // dataset.metas[0].type = 'tabbed' and dataset.metas[1] = 'line' if chosen UserWhiteNoise
+      // dataset.metas[0] - 'line' if Dummy chosen
+      if (dataset.metas[0].type === 'tabbed') {
+        this.datasetPref = (dataset.metas[0] as any).metas[0];
+      } else if (dataset.metas[0].type === 'line') {
+        this.datasetPref = dataset.metas[0] as LineChartMeta;
+      }
+      // calling this.init() from inside bc when code above is in constructor,
```
mfitz3, line 60: Can probably remove this comment, and rename `init` to something more specific.

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
```diff
@@ -1,32 +1,88 @@
...
+      // and code in this.init() is in ngOnInit() there is sync problem
+      this.init();
+    });
   }
 
   ngOnDestroy() {
```
mfitz3, line 66: Can remove empty methods.

src/d3/scatterplot.d3.spec.ts
```diff
@@ -43,45 +57,51 @@ describe('VR Scatter Plot', () => {
...
   });
+
   it('Check for yGrid', () => {
-    scatterplot.init(element, []);
-    const yGrid = document.getElementById('yGrid');
-    let yGridPresent = false;
-    if (yGrid !== null){
-      yGridPresent = true;
+    let yGridPresent: boolean = false;
+    for (let i = 0; i < grids.length; i++){
```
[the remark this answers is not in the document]
cvillapudua: I didn't know about this function, but will def help with conciseness of code so thanks!

src/d3/scatterplot.d3.spec.ts
```diff
@@ -1,40 +1,54 @@
...
+
+  // let scatterPlotData1: VRTimeSeriesPoint[]  = [];
+  // scatterPlotData1.push({x: new Date(), y: 10, z: 10});
+
+  let lineChartData8: TimeSeriesPoint[] = [];
+  for (let i = 0; i < 8; i++){
+    lineChartData8.push({x: new Date(), y: i * 10});
+  }
+  
+  // let scatterPlotData8: VRTimeSeriesPoint[] = [];
```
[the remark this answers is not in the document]
cvillapudua: I was planning on having the commented code set up testing once I merge `vrDataset` branch with the result of this PR merge (PR #64). If it's best practice to just delete all commented code, I can go ahead and do that as well. Lmk what you think is best.

src/d3/scatterplot.d3.spec.ts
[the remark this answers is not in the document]
cvillapudua: @mfitz3 just realized `.some()` is only for simple arrays, but `grids` is an `HTMLCollectionOf<Element>`. It gives me an error when I try to cast to an array of `Elements`. Also, with `.some()` the only thing passed in the element and I can't specify which specific `id` I need, whether that be `xGrid`, `yGrid`, or `zGrid`. 

I'm thinking of solutions... One could be passing in `grids` and the `id` I'm searching for and if that `id` is found in one of the elements of `grids` I can return true. I don't know if there are any existing functions that are similar (`HTMLCollectionOf` only has two functions and neither seem to do what I need for this).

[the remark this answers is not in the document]
mfitz3: You can use `Array.from`:

`Array.from(grides).some(...)`

src/d3/scatterplot.d3.spec.ts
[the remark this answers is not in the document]
mfitz3: Clearing the commented code is usually ideal, just to produce cleaner PRs for when people are looking at the change history.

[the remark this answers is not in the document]
cvillapudua: @mfitz I went ahead and removed them so that this PR can just contain working code!

src/components/vr-scatter-plot/vr-scatter-plot.component.ts
cvillapudua: went ahead and removed comment and renamed `.init()` to `.setDataOnD3()`

src/d3/scatterplot.d3.spec.ts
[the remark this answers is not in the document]
cvillapudua: @mfitz `some` only allows me to check one condition per function so I'd have to write 3 functions to check for each of the `id` cases individually. Would you prefer I go with the solution I mentioned above or use `some` and write those 3 checking functions?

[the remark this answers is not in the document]
cvillapudua:
```it('GridHelper xGrid', () => {
    expect(findGrids(grids, 'xGrid')).toEqual(true);
  });

  it('GridHelper yGrid', () => {
    expect(findGrids(grids, 'yGrid')).toEqual(true);
  });

  it('GridHelper zGrid', () => {
    expect(findGrids(grids, 'zGrid')).toEqual(true);
  });
});

function findGrids(entities: HTMLCollectionOf<Element>, gridID: string): boolean{
  for (const entity of (entities as any)){
    if (entity.id === gridID){
      return true;
    }
  }
  return false;
}``` what the solution looks like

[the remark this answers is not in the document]
cvillapudua: this is the solution I wrote, but can change to using `some` if that would be better... The method above saves lines of code and is a bit more concise than writing three different checking functions, but it seems that both ways are similar in execution time

PR #59, opened 2020-07-15
Accessibility VR Page

HarrisonPearl:
Added a new page for accessible VR Features, with the goal of supporting blind/visually impaired users.

Changes include:

 - a VR Accessibility Component with an embedded A-Frame VR element and support for controller interactions
 - a D3 document, hapticplot.d3, which generates A-frame objects and performed necessary event listener set-up and element 
    styling
 - angular routing changes to support navigation to the new component
 - three new dependencies (`aframe-extras`, `super-hands`, and `aframe-haptics-component`) to support controller interactions


src/app-routing.module.ts
```diff
@@ -3,18 +3,22 @@ import { RouterModule, Routes } from '@angular/router';
...
+  { path: 'accessibilityVR', component: VRAccessibilityComponent, pathMatch: 'full' },
 ];
 
 @NgModule({
   imports: [
     RouterModule.forRoot(routes),
     DashboardModule,
-    VRScatterPlotModule
+    VRScatterPlotModule,
+    VRAccessibilityModule
```
danielyule, line 21: World's biggest nit: it's slightly nicer to put a trailing comma here so that in the next diff, the only line that's changed is the one that is added.

src/components/vr-accessibility/vr-accessibility.component.html
```diff
@@ -0,0 +1,13 @@
...
+-->
+<a-scene embedded style="position: absolute; height: 90vh;">
+    <a-assets></a-assets>
+    <a-entity>
+      <a-camera></a-camera>
+      <a-entity sphere-collider super-hands oculus-touch-controls="hand: left;" haptics="events: triggerdown; dur: 1000; force: 0.5"></a-entity>
+      <a-entity sphere-collider super-hands oculus-touch-controls="hand: right;" haptics="events: triggerdown; dur: 1000; force: 0.5"></a-entity>
+    </a-entity>
+    <a-box hoverable grabbable stretchable draggable dropppable color="green" position="-1 0 -1"></a-box>
+  </a-scene>
```
danielyule, line 13: Ultra nit: the indent spacing isn't quite right here.

src/components/vr-accessibility/vr-accessibility.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  templateUrl: './vr-accessibility.component.html'
+})
+export class VRAccessibilityComponent implements OnInit, OnChanges, OnDestroy{
+  private vrHapticPlot: Hapticplot;
+  private shape: string;
+  private color: string;
+
+  constructor(
+  ) {
+    this.vrHapticPlot = new Hapticplot('a-sphere');
```
danielyule, line 21: You can put this on line 15 where the property is declared. Empty constructors are best constructors.

src/components/vr-accessibility/vr-accessibility.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+  ngOnInit() {
+   this.vrHapticPlot.init(document.querySelector('a-scene'), [0, 1/5, 2/5, 3/5, 4/5, 5/5, 6/5]);
```
danielyule, line 24: I think you can use a [ViewChild](https://blog.angular-university.io/angular-viewchild/) to access this, rather than a query selector.

src/components/vr-accessibility/vr-accessibility.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  ngOnDestroy() {
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    // if ('measureName' in changes) {
```
danielyule, line 31: Might be an artefact of the broken testing, but never check in commented code.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+describe('VR Haptic Plot', () => {
+  const shape = 'a-sphere';
+  let element: HTMLElement;
+  let hapticplot: Hapticplot;
+
+  beforeEach( () =>  {
+    element = document.createElement('a-scene');
+    hapticplot = new Hapticplot(shape);
+  });
+  it('places no points bc 1:1 correspondence with empty element array', () => {
```
danielyule, line 13: Super nit: prefer spaces between test cases.

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
+// run `tsc Hapticplot.ts` to compile into Hapticplot.js file
+
+// for running on browser
+// import * as _d3 from 'd3';
+
+// declare global {
+// const d3: typeof _d3;
+// }
```
danielyule, line 8: Are these instructions still necessary?

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
...
+
+// for running on unit tests
+import * as d3 from 'd3';
+import * as THREE from 'three';
+
+export class Hapticplot{
+    private data: number[];
+    private shape: string;
+    private container: HTMLElement | null;
```
danielyule, line 17: Should be initialized to null or made optional.  Either 
```typescript
private container: HTMLElement | null = null;
```
or 
```typescript
private container?: HTMLElement;
```

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
...
+  constructor(shape: string) {
+    this.shape = shape;
+  }
+  init(container: HTMLElement | null, data: number[]){
+    this.data = data;
+    this.container = container;
+    // create a scale so that there is correspondence between data set and screen render
+    this.hscale = d3.scaleLinear();
+    this.hscale.domain([0, d3.max(this.data) as number])       // max of dataset
+    .range([0, 100]);                                      // linear mapping of data set values to values from 0 to 10
```
danielyule, line 29: Should be indented to show it's the continuation of the previous line

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.createGridPlane();
+  }
+
+  private generatePts() {
+     /*
+     Generates points based on initilization data
+      - "enter" identifies any DOM elements to be added when # array elements & # DOM elements don't match
+      - select all generated entities and updates their position based on ingested data
+    */
+    d3.select(this.container).selectAll(this.shape).data(this.data).enter().append(this.shape);
```
danielyule, line 45: Typically prefer selecting on css class, rather than element name.  It's been shown to be much faster.

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
...
+      const y = this.data[i];
+      const z = -1;
+      return `${x} ${y} ${z}`;
+    });
+  }
+  private setColor(color) {
+    /*
+    Adds given color property to all generated entities
+    */
+    d3.select(this.container).selectAll(this.shape).attr('color', () => {
```
danielyule, line 57: Seems like you could make a utility method that does this selectAll thing, since you are doing it so much

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.hscale = d3.scaleLinear();
+    this.hscale.domain([0, d3.max(this.data) as number])       // max of dataset
+    .range([0, 100]);                                      // linear mapping of data set values to values from 0 to 10
+    this.createSky();
+    this.generatePts();
+    this.setColor('blue');
+    this.setRadius(0.1);
+    this.setInteraction();
+    this.addEventListeners();
+    this.createGridPlane();
```
danielyule, line 36: Rather than having a method for each piece of the d3 initialization, it's standard to just do it all in one cascade, since they rarely need to be updated independently. That way when the data updates later, you don't have to go through and call all the methods again.

What I mean by one big cascade is like
```typescript
d3.select(this.container).selectAll(this.shape)
  .attr('position', position)
  .attr('radius', radius)
  .attr('color', color);
```
And so on.

If you do end up doing that, then you typically wouldn't put things like callback functions inline; you'd define them elsewhere.

src/components/vr-accessibility/vr-accessibility.component.ts
nattySP: If you are able to use ViewChild, this line will have to be moved into an ngAfterViewInit()

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
...
+export class Hapticplot{
+    private data: number[];
+    private shape: string;
+    private container: HTMLElement | null;
+    private hscale: d3.ScaleLinear<number, number>;
+
+  constructor(shape: string) {
+    this.shape = shape;
+  }
+  init(container: HTMLElement | null, data: number[]){
```
nattySP, line 23: nit: whitespace between methods

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
...
+      .attr('draggable', '')
+      .attr('dropppable', '');
+
+  }
+
+  private addEventListeners() {
+    /*
+      adds event listeners to react if a generated data object is hovered or un-hovered
+    */
+    d3.select(this.container).selectAll(this.shape).on('stateadded', function(d, i) {
```
nattySP, line 87: I would declare a local variable for the shape selection 
e.g. const shapes = d3.select(this.container).selectAll(this.shape)
The idea is that reading a name for concept (aka a local variable) creates less cognitive overhead than mentally parsing 'd3.select(this.container).selectAll(this.shape)';  or if you take Daniel's suggestion to create a util method, it accomplishes the same thing -- two birds one stone. 
e.g. this.getShapes().on('stateadded'....)

src/d3/hapticplot.d3.ts
```diff
@@ -0,0 +1,137 @@
...
+    d3.select(this.container).selectAll(this.shape).on('stateremoved', function(d, i) {
+      if (d3.event.detail === 'hovered'){
+        d3.select(this).attr('color', 'blue');
+      }
+    });
+  }
+
+
+
+  private createSky(){
```
nattySP, line 101: It looks like createSky and createGridPlane are both repeated in hapticplot and scatterplot; can you DRY this up?  Maybe a helper class that takes care of creating these things and can be used by both.

src/components/vr-accessibility/vr-accessibility.component.ts
HarrisonPearl: switched it over, and added ngAfterViewInit()

src/d3/hapticplot.d3.ts
HarrisonPearl: nope!

src/d3/hapticplot.d3.ts
HarrisonPearl: I switched the selectAll method to use the css class "datapoint," but I'm having issues when I try to switch the container to a css selection

src/d3/hapticplot.d3.ts
HarrisonPearl: done! but I also condensed the functions that were using it in to one function that only does it once, so this might no longer be necessary. might be nice to leave in for readability though.

src/d3/hapticplot.d3.ts
HarrisonPearl: I condensed them all into one cascading function! I'm having a bit of trouble defining the callbacks outside of it though. For the first call back, (which sets the objects positions) I need access to the (d, i) vales from the d3 selection, but I was able to pull the logic out into a helper function called "generatePositions." For the event listeners callback, however, I need to maintain access to the event and the currently selected entity, which I'm not sure how to maintain when passing it to the helper.

src/d3/hapticplot.d3.ts
HarrisonPearl: Good call! I'm planning on deviating my implementation of createGridPlanes significantly, so I think it will be best to keep that one seperate. createSky is pretty generic though, I'll work on creating a helper class and talk with carlos about whether what else we can do to avoid redundancies.

src/d3/hapticplot.d3.ts
danielyule: I should have been more specific; I just meant the shape.  The container can stay as it is, since it's already got the element itself

PR #63, opened 2020-07-16
Add raw city data and a binary to refine them

64json:
Run `bin/refine_city_data` would do the following:
1. Fetch country data.
2. Read city data and convert CSV to JSON.
3. Filter out unnecessary data to reduce size.
4. Combine city and country data.
5. Write the combined data to `src/assets/cities.json`.


jcc37:
Is that the correct way to include the CC4.0 database into our repository under the Apache license? I am not an expert in law but I want to be careful about these license things.

I think `src/assets/cities.json` can be considered as a modified version and redistribution of the database, which should also follow the terms of attribution. I would suggest putting information on README indicating the use of this database if needed.

64json:
> Is that the correct way to include the CC4.0 database into our repository under the Apache license? I am not an expert in law but I want to be careful about these license things.
> 
> I think `src/assets/cities.json` can be considered as a modified version and redistribution of the database, which should also follow the terms of attribution. I would suggest putting information on README indicating the use of this database if needed.

The `license.txt` says "this [https://simplemaps.com/data/us-cities] backlink must come from a **public webpage where the Customer is using the data**," and CC BY 4.0 states that "you must give appropriate credit, provide a link to the license, and **indicate if changes were made**."

So I am planning to include the attribution on the UI by the time I mark this PR as ready for review, but not on the README.

Issue #54, opened 2020-07-10
Refactor unit tests


jcc37:
In PR #46 , old tests are fixed and updated for the new data flow with dataset support. However, according to the comments from @Rhombus33 and @jnager , several code refactoring and changes are necessary to our current tests. These changes should implement the following items and be completed in one or more PRs:

- Test the user-facing behavior instead of part of the implementation (i.e. if the tabs div is present in the DOM)
- Assign literal values and anonymous functions to constants
- Separate tests of a component with different type variable parameters to distinct describe blocks
- Tests should import modules rather than declare components when setting up TestBed.
  - Should we rename `xxx.component.spec.ts` to `xxx.module.spec.ts` since we are now importing the entire module when testing?

Before this issue gets resolved, tests should still be added and updated to each PR along with any new features or changes in the data flow.

64json:
- Mocked dependencies: https://github.com/googleinterns/guide-doge/pull/46#discussion_r452382101

64json:
- Arrange-Act-Assert structure: https://github.com/googleinterns/guide-doge/pull/46#discussion_r452377529

PR #67, opened 2020-07-19
Fix duplicate dataset creation

### LINKED ISSUE #66: Create dataset twice when changing dataset
The dataset is created twice with the same set of configs when changing the dataset, which makes the charts rendered twice with a blink.

It happens when changing dataset from A to B where configMeta of B is equal to or a subset of configMeta of A.

jcc37:
Resolve #66 

Tests for this bug should be added when resolving #54


src/services/data/data.service.ts
```diff
@@ -52,15 +37,17 @@ export class DataService implements OnDestroy {
...
-            options: Object.keys(datasets),
+            defaultValue: defaultDatasetName,
+            options: datasetNames,
           },
           ...datasets[name].configMeta,
         };
 
         this.preferenceService.dataset$.next({
           ...createDefault(meta),
+          name,
```
Rhombus33, line 47: Why does there need to be 2 properties with the same value?

src/services/data/data.service.ts
```diff
@@ -52,15 +37,17 @@ export class DataService implements OnDestroy {
           },
           name: {
             type: 'select',
-            defaultValue: name,
-            options: Object.keys(datasets),
+            defaultValue: defaultDatasetName,
```
Rhombus33, line 39: Why do we need to define the default for each one if it is always the same value?

jcc37: We need to make new meta object contain `fields name` and `enabled`.  It is going to replace the meta object of previous dataset.

src/services/data/data.service.ts
jcc37: There are two subscription in `DataService`: One for updating meta and one for creating data when preference changes. When user changes the `name` in preference, the `_meta_name` does not change, so the subscription above will not be triggered to create new data. After `_meta_name` is changed, which means the meta in preference is update to date, that subscription will then be triggered.

https://github.com/googleinterns/guide-doge/pull/67/files/09a2e79376e508f458b5aa82e4446a43bd5a1c03#diff-04c7fa6e46d3da32b3b6976c27dd53edR19

PR #69, opened 2020-07-21
adds haptic triggers to data entities

HarrisonPearl:
Adds a d3 function to set up haptic triggers on generated data objects. 
 - Replaces `modifyOnStateChange` function with `onHoverStart` and `onHoverEnd`
 - Adds triggers to fire haptic pulses when a `hover-start` event is emitted by a data object
 - Accesses the controller entity through the `event.details.hand` property
 - Initializes controller entities' haptic component force property with value 0 (when the `.pulse` method is called, it triggers a pulse with the initialization force value + the passed value, and the default initialization is the max. Setting it to 0 allows the passed value to control the haptics)

### DIFF
--- a/src/components/vr-accessibility/vr-accessibility.component.ts
+++ b/src/components/vr-accessibility/vr-accessibility.component.ts
@@ -32,7 +32,7 @@ export class VRAccessibilityComponent implements OnInit, OnChanges, OnDestroy, A
 
   ngAfterViewInit(){
     const scene = this.theScene.nativeElement;
-    this.vrHapticPlot.init(scene, [0, 1 / 5, 2 / 5, 3 / 5, 4 / 5, 5 / 5, 6 / 5]);
+    this.vrHapticPlot.init(scene, [0, 0, 0.5, 0.5, 1, 1]);
   }
 
 }
--- a/src/d3/hapticplot.d3.spec.ts
+++ b/src/d3/hapticplot.d3.spec.ts
@@ -65,7 +65,7 @@ describe('VR Haptic Plot', () => {
 
 // Helper Functions
 
-// returns array of actual position vectors
+// Returns an array of actual position vectors
 function getPosition(scene: HTMLElement, shape: string): Vector3[]{
   const attrArray: Vector3[] = [];
   Array.from(scene.querySelectorAll(shape)).forEach((child) => {
@@ -77,25 +77,26 @@ function getPosition(scene: HTMLElement, shape: string): Vector3[]{
   return attrArray;
 }
 
-// returns array of each generated objects color
+// Returns an array of each generated objects color
 function getColor(scene: HTMLElement, shape: string): (string | null)[]{
   return Array.from(scene.querySelectorAll(shape)).map((point: Element) => point.getAttribute('color'));
 }
 
-// returns array of each generated objects radius
+// Returns an array of each generated objects radius
 function getRadius(scene: HTMLElement, shape: string): (string | null)[]{
   return Array.from(scene.querySelectorAll(shape)).map((point: Element) => point.getAttribute('radius'));
 }
 
-// returns array of each generated objects hover property
+// Returns an array of each generated objects hover property
 function getHoverable(scene: HTMLElement, shape: string): (string | undefined)[]{
-  const attrArray: (string | undefined)[] = [];
-  Array.from(scene.querySelectorAll(shape)).forEach((child) => { attrArray.push((child as Entity).components.hoverable.attrName); });
-  return attrArray;
+  return Array.from(scene.querySelectorAll(shape)).map((point: Element) => (point as Entity).components.hoverable.attrName);
 }
 
-// returns array of each generated objects color, after a hover event has occured
+// Returns an array of each generated objects color, after a hover event has occured
 function getHoveredColor(scene: HTMLElement, shape: string): (string | null)[]{
-  Array.from(scene.querySelectorAll(shape)).forEach((child) => { child.dispatchEvent(new Event('hover-start')); });
+  const shapes = scene.querySelectorAll(shape);
+  for (const point of Array.from(shapes)){
+    point.dispatchEvent(new Event('hover-start'));
+  }
   return Array.from(scene.querySelectorAll(shape)).map((point: Element) => point.getAttribute('color'));
 }
--- a/src/d3/hapticplot.d3.spec.ts
+++ b/src/d3/hapticplot.d3.spec.ts
@@ -1,5 +1,6 @@
 import { Hapticplot } from './hapticplot.d3';
 import { Entity } from 'aframe';
+import { Vector3 } from 'three';
 
 
 describe('VR Haptic Plot', () => {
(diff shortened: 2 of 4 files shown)


src/d3/hapticplot.d3.ts
```diff
@@ -14,17 +14,21 @@ export class Hapticplot{
...
     this.data = data;
     this.container = container;
-    // create a scale so that there is correspondence between data set and screen render
+    // create a scale so that there is correspondence between data set and screen render,
+    // a linear mapping of data set values to values from 0 to 10
     this.hscale = d3.scaleLinear();
     this.hscale.domain([0, d3.max(this.data) as number])  // max of dataset
       .range([0, 100]);
-    this.setupPoints('blue', 0.1);                        // linear mapping of data set values to values from 0 to 10
+    this.setupPoints('green', 0.05);
```
nattySP, line 22: This is probably stylistic preference but it might make sense to make this value a file level constant like
`const POINT_SIZE = 0.05`; tends to be more maintainable.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -53,3 +87,44 @@ function getPosition(element: HTMLElement, shape: string): Array<{x: number, y:
...
 }
+
+// returns array of each generated objects color
+function getColor(element: HTMLElement, shape: string): Array<string>{
+  const childrenArray = element.querySelectorAll(shape);
+  const attrArray: Array<string> = [];
+  for (const child of (childrenArray as any)){
+    attrArray.push((child as any).getAttribute('color'));
+  }
+  return attrArray;
```
nattySP, line 98: Something cool you can do is 
```suggestion
  return Array.from(element.querySelectorAll(shape).map((point: 'type of point') => point.getAttribute('color'))
```
This is more idiomatic typescript; also avoid using `any`....I'm not sure what 'type of point' is but if you need help correctly annotating this let me know!
Same comment applies to all these helper methods.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -42,8 +42,42 @@ describe('VR Haptic Plot', () => {
...
+    const result = getHoveredColor(element, shape);
+    expect(result).toEqual(expectedSizeArray);
+  });
 });
 
+/*
+*
+Helper Functions
+*
+*/
```
danielyule, line 79: Should probably be a one line comment, or at least have consistent asterixes

src/d3/hapticplot.d3.spec.ts
```diff
@@ -53,3 +87,44 @@ function getPosition(element: HTMLElement, shape: string): Array<{x: number, y:
   }
   return positionArray;
 }
+
+// returns array of each generated objects color
+function getColor(element: HTMLElement, shape: string): Array<string>{
+  const childrenArray = element.querySelectorAll(shape);
+  const attrArray: Array<string> = [];
```
danielyule, line 94: The TS styleguide suggests string[] instead of Array<string> for non union types

src/d3/hapticplot.d3.spec.ts
```diff
@@ -53,3 +87,44 @@ function getPosition(element: HTMLElement, shape: string): Array<{x: number, y:
   }
   return positionArray;
 }
+
+// returns array of each generated objects color
+function getColor(element: HTMLElement, shape: string): Array<string>{
+  const childrenArray = element.querySelectorAll(shape);
```
danielyule, line 93: as SVGElement[] would prevent the casts to `any` later on.  In general, avoid `any` at all costs (you can use `unknown` where it makes sense, or cast it to what you need)

danielyule: Or do what Natalie suggested, that's even better

src/d3/hapticplot.d3.ts
```diff
@@ -46,31 +50,43 @@ export class Hapticplot{
...
 
+  private onHoverStart(entity, hapticIntensity, color){
+    /*
+    When an object begins being hovered by the controller entity
+     - triggers a haptic pulse
+     - changes the entities color to indicate a pulse has fired
+    */
+    if (d3.event.detail !== undefined && d3.event.detail.hand !== undefined){
+      d3.event.detail.hand.components.haptics.pulse(hapticIntensity, 1000);
+    }
```
danielyule, line 78: TypeScript has a really neat new operator for [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining): `?.`.

You can re-write this as `d3.event.detail?.hand?.haptics.pulse(hapticIntensity, 1000);`

src/d3/hapticplot.d3.ts
```diff
@@ -46,31 +50,43 @@ export class Hapticplot{
...
-    if (d3.event.detail === state){
-      d3.select(entity).attr('color', color);
-    }
-  }
-
-  // Generates a world space position for each data entity, based on ingested data
   private generatePositions(data, index){
+    /*
+      Generates a world space position for each data entity, based on ingested data
+    */
```
danielyule, line 63: I think this should be a JSDoc comment (use two asterixes and put it outside the method). Same elsewhere in this file.

src/d3/hapticplot.d3.ts
```diff
@@ -46,31 +50,43 @@ export class Hapticplot{
...
     const x = index / 10;
     const y = data;
     const z = -1;
     return `${x} ${y} ${z}`;
   }
 
+  private onHoverStart(entity, hapticIntensity, color){
```
danielyule, line 70: Not sure color needs to be passed in here; seems more like a constant. Same for hover end

src/components/vr-accessibility/vr-accessibility.component.ts
```diff
@@ -32,7 +32,7 @@ export class VRAccessibilityComponent implements OnInit, OnChanges, OnDestroy, A
 
   ngAfterViewInit(){
     const scene = this.theScene.nativeElement;
-    this.vrHapticPlot.init(scene, [0, 1 / 5, 2 / 5, 3 / 5, 4 / 5, 5 / 5, 6 / 5]);
+    this.vrHapticPlot.init(scene, [0, 0, 0.5, 0.5, 1, 1]);
```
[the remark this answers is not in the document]
HarrisonPearl: This is being used to represent the data from which the data entities' positions and haptics are being generated. Right now it is just dummy values for testing, but the plan is to switch to loading the datasets being used throughout the project dynamically. I'll add a comment to the code to provide details.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -42,8 +42,42 @@ describe('VR Haptic Plot', () => {
     const result = getPosition(element, shape);
     expect(result).toEqual(expectedPosArray);
   });
+
+  it('places points for each element in a three element array, and checks their color property', () => {
```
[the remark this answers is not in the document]
HarrisonPearl: I gave the HTMLElement variable a more descriptive name, and replaced all mentions of the word element with datum in the test descriptions to distinguish between them. I think its much more clear now!

src/d3/hapticplot.d3.ts
HarrisonPearl: I added `POINT_SIZE` `DEFAULT_COLOR` and `HOVER_COLOR` as file level constants in the init function, and updated their use throughout.

src/d3/hapticplot.d3.spec.ts
HarrisonPearl: done!

src/d3/hapticplot.d3.ts
HarrisonPearl: I switched them over!

src/d3/hapticplot.d3.spec.ts
```diff
@@ -1,55 +1,101 @@
...
+// returns array of each generated objects radius
+function getRadius(scene: HTMLElement, shape: string): (string | null)[]{
+  return Array.from(scene.querySelectorAll(shape)).map((point: Element) => point.getAttribute('radius'));
+}
+
+// returns array of each generated objects hover property
+function getHoverable(scene: HTMLElement, shape: string): (string | undefined)[]{
+  const attrArray: (string | undefined)[] = [];
+  Array.from(scene.querySelectorAll(shape)).forEach((child) => { attrArray.push((child as Entity).components.hoverable.attrName); });
+  return attrArray;
```
danielyule, line 94: This should probably be a map and not a forEach.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -1,55 +1,101 @@
...
+}
+
+// returns array of each generated objects color, after a hover event has occured
+function getHoveredColor(scene: HTMLElement, shape: string): (string | null)[]{
+  Array.from(scene.querySelectorAll(shape)).forEach((child) => { child.dispatchEvent(new Event('hover-start')); });
```
danielyule, line 99: This line would probably be easier to read as a for..of loop.  Google's styleguide actually forbids `forEach` entirely, but in this case in particular it would dodge the `Array.from`

src/d3/hapticplot.d3.ts
```diff
@@ -12,81 +13,108 @@ export class Hapticplot{
...
   }
 
-  // Generates a world space position for each data entity, based on ingested data
-  private generatePositions(data, index){
+  /**
+   * Generates a world space position for each data entity, based on ingested data
+   * @param data Ingested Data
+   * @param index Crrent index in data array
+   */
+  private generatePositions(data, index, entity){
```
danielyule, line 72: With what this actually does now, it should probably be `setPosition`.

src/d3/hapticplot.d3.ts
```diff
@@ -12,81 +13,108 @@ export class Hapticplot{
...
+   *  - changes the entities color to indicate hovering has ended
+   * @param entity The data entity which is no longer hovered
+   * @param defaultColor The default color of unhovered entities
+   */
+  private onHoverEnd(entity, defaultColor){
+    d3.select(entity).attr('color', defaultColor);
   }
 
+  /**
+   * creates and adds a sky box to the scene
```
danielyule, line 103: Nit: could be one line and also should be capitalized.

src/d3/hapticplot.d3.spec.ts
HarrisonPearl: I changed it to a for...of loop, but I'm still using Array.from since the NodeListOf<Element> returned by the querySelector has no iterator. Is there a better way to do this?

PR #76, opened 2020-07-30
Fix undeterministic test failure

64json:
Found out it was caused by one of the tests feeding `'INVALID_DATASET_NAME'` as dataset name.

The test has always been throwing an error, but Jasmine ignored the error depending on the randomized order of running tests, which made it undeterministically succeeded.

https://github.com/googleinterns/guide-doge/blob/3a79f33fad0066751b90477b8b0d5fe7ca0d6c34/src/services/data/data.service.spec.ts#L45

### DIFF
--- a/src/services/data/data.service.ts
+++ b/src/services/data/data.service.ts
@@ -45,6 +45,7 @@ export class DataService implements OnDestroy {
       .pipe(distinctUntilChanged((prev, curr) => prev.name === curr.name))
       .pipe(pluck('name'))
       .pipe(map(name => name ?? Object.keys(datasets)[0]))
+      .pipe(filter(name => name in datasets))
       .subscribe(name => {
         const meta = {
           enabled: {


jcc37:
I think this issue is more important than we thought before. This is a bug which should be "always" detected by this test, but as you said the test turned out to random succeed (even most of the time). It means this test is not work properly and we should also fix this bug of test.

I guess it is related to the microtask of `observable.subscribe`, the `expect` is executed before the subscription microtasks are flushed so it takes the old value of observable. If this is the case, other tests of our services may also have this unintentional success issue.

64json:
> I guess it is related to the microtask of `observable.subscribe`, the `expect` is executed before the subscription microtasks are flushed so it takes the old value of observable. If this is the case, other tests of our services may also have this unintentional success issue.

iirc and debugged it correctly, it never used the old value of observable, and the subscription callback always threw an error.

I think it is because the subscription happens within `beforeEach`. By the time the subscription callback throws an error, it doesn't propagate back to `beforeEach` and just gets lost.

PR #78, opened 2020-07-31
Adds controller mock and haptic initialization testing

HarrisonPearl:
Adds a mock controller to the testing setup, and attaches relevant components. 
- Implements a `getHaptics` helper function to check that the haptic component attaches successfully to the controller mock.
- Removes use of forEach in the `getPositions` helper function.
- Renames helper functions to better reflect their effects.

### DIFF
--- a/src/d3/hapticplot.d3.spec.ts
+++ b/src/d3/hapticplot.d3.spec.ts
@@ -75,7 +75,7 @@ describe('VR Haptic Plot', () => {
 
   it('initilizes the controller entitys and attaches the haptic component', () => {
     const expectedRes = true;
-    const result = getHaptics(controller);
+    const result = hasHaptics(controller);
     expect(result).toEqual(expectedRes);
   });
 });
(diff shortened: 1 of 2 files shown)


src/d3/hapticplot.d3.spec.ts
```diff
@@ -1,89 +1,107 @@
...
     attrArray.push(new Vector3(
-      (child as Entity).object3D.position.x,
-      (child as Entity).object3D.position.y,
-      (child as Entity).object3D.position.z));
-  });
+      (point as Entity).object3D.position.x,
+      (point as Entity).object3D.position.y,
+      (point as Entity).object3D.position.z));
+  }
   return attrArray;
```
danielyule, line 95: I think you can turn this whole thing into a map:

```typescript
return scene.querySelectorAll(shape)
  .map((child: Entity) => 
    new Vector3(
      point.object3D.position.x,
      point.object3D.position.y,
      point.object3D.position.z));
```
I'm not 100% on if the type will work in the function signature, but if it doesn't, you can just use the cast like you have been

PR #80, opened 2020-08-03
Trend protoform and regression text summarization

jcc37:
**Note: Trend protoform (`trend.summarization.ts`) has been deprecated, and trend regression (`trend-regression.summarization.ts`) is included in the current plan under development but changed a lot**
- Add summarization modules `Trend` (naive protoform) and `TrendRegression`
- Add new dataset `UserWorkdayHolidayLinear`
- Tests for new modules and datasets

Not included in this PR:
- Tests for trend cores (regression, partially linear approximation functions): We should develop a test suite for testing all core functions


src/datasets/dummy.dataset.ts
```diff
@@ -24,7 +25,7 @@ export function create(config: Config): Dataset {
   for (let i = 1; i <= 100; i++) {
     data.push({
       x: new Date(time.getTime() + i * DAY),
-      y: Math.exp(i / 10) + config.offset + rand(),
+      y: Math.exp(i / 10) + config.offset + rand() / 2,
```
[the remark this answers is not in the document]
jcc37: thanks for pointing out. The variables for these constants had been added to other datasets in previous PRs and it seems to be a miss for this dataset.

src/datasets/summarizations/libs/trend.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+export function linearRegression(points: NumPoint[]): LinearRegressionResult {
+  const pairs = points.map(pointToPair);
+  const result = regression.linear(pairs);
+  const gradient = result.equation[0];
+  const gradientAngleRad = Math.atan(gradient);
+  const prediction = result.points.map(pairToPoint);
+
+  const errors = points.map(({ y }, i) => {
```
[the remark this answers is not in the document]
jcc37: Maybe using `point.y` here is more readable. It will become`math.abs(point.y - prediction[i].y)`

src/datasets/summarizations/libs/trend.ts
```diff
@@ -0,0 +1,127 @@
...
+  idxStart: number;
+  idxEnd: number;
+  timeStart: Date;
+  timeEnd: Date;
+  pctSpan: number;
+  cone: Cone2D;
+}
+
+
+export function normalizedUniformPartiallyLinearEpsApprox(points: TimeSeriesPoint[], eps: number): TimeSeriesPartialTrend[] {
```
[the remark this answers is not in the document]
jcc37: It is the notation used in the paper. I will add a reference to the paper in the comments.

src/datasets/summarizations/trend-regression.summarization.ts
```diff
@@ -0,0 +1,139 @@
...
+          overallLinearTrendValidity,
+          uLinearTrend(overallLinearModel.gradientAngleRad),
+          weekdayLinearTrendValidity,
+          uLinearTrend(weekdayLinearModel.gradientAngleRad),
+          weekendLinearTrendValidity,
+          uLinearTrend(weekendLinearModel.gradientAngleRad),
+        )),
+    );
+
+    for (const [linearTrend, uLinearTrend] of uLinearTrends) {
```
Rhombus33, line 98: Can you add a small comment for each of these for blocks? They are all very similar and it takes a bit to figure out what summary each is for.

src/datasets/summarizations/trend-regression.summarization.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { queryFactory } from './trend-regression.summarization';
+import { TimeSeriesPoint } from '../metas/types';
+
+describe('queryFactory', () => {
```
Rhombus33, line 4: Could we add a couple tests for the summaries generated? Would be nice to use a couple clean/simple datasets to sanity check that the summaries output are correct. (i.e. the current linearly increases dataset should always generate the same summaries).

src/datasets/summarizations/trend-regression.summarization.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    const uWeekend = (p: TimeSeriesPoint) => p.x.getDay() === 5 ? 0.2 : +(p.x.getDay() === 0 || p.x.getDay() === 6);
+    const uWeekday = (p: TimeSeriesPoint) => 1 - uWeekend(p);
+
+    const isWeekend = (p: TimeSeriesPoint) => uWeekend(p) > 0.5;
+    const isWeekday = (p: TimeSeriesPoint) => uWeekday(p) > 0.5;
+
+    const weekdayWeekendRatioPoints = groupPointsByXWeek(points).map(weekPoints => {
+      const week = weekPoints[0].x;
+      const wWeekdays = math.sum(weekPoints.map(uWeekday));
```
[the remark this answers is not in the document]
jcc37: Actually I don't know how to call these two terms. They are numerator and denominator to compute the weighted average.

src/datasets/summarizations/trend-regression.summarization.spec.ts
jcc37: I added two simple test data to verify the created summaries and validity. Let me know your thoughts about it!

src/datasets/dummy.dataset.ts
```diff
@@ -18,13 +19,17 @@ export const configMeta: PreferenceMeta<Config> = {
 };
 
 export function create(config: Config): Dataset {
-  const time = new Date(Date.now() - 100 * DAY);
+  const startDaysBeforeNow = 100;
```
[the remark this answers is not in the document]
jcc37: How about `pointsLength`? I would like to rename `data` to `points` also.

src/datasets/summarizations/trend-regression.summarization.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    expect(otherSummaries.some(isValidSummary)).toBeFalse();
+  });
+
+  it('should create summaries describing weekday and weekend linear trends.', () => {
+    const summaries = queryFactory(weekdayQuicklyIncreasingWeekendConstantPoints)();
+    const isValidSummary = ({ validity }) => validity >= 0.7;
+    const isWeekdayQuicklyIncreasingSummary = ({ text }) => text.includes('weekday') && text.includes('linearly quickly increasing');
+    const isWeekendConstantSummary = ({ text }) => text.includes('weekend') && text.includes('constant');
+
+
```
Rhombus33, line 62: nit: Remove extra line

src/datasets/dummy.dataset.ts
```diff
@@ -18,13 +19,17 @@ export const configMeta: PreferenceMeta<Config> = {
 };
 
 export function create(config: Config): Dataset {
```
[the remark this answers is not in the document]
jcc37: Added. Actually, I am thinking to have a description field in the dataset object and we can display it to users in a popup of help icon.

src/datasets/summarizations/trend-regression.summarization.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    const summaries = queryFactory(overallQuicklyIncreasingPoints)();
+    const isValidSummary = ({ validity }) => validity >= 0.7;
+    const isOverallQuicklyIncreasingSummary = ({ text }) => text.includes('overall') && text.includes('linearly quickly increasing');
+
+    const overallQuicklyIncreasingSummaries = summaries.filter(isOverallQuicklyIncreasingSummary);
+    const otherSummaries = summaries.filter(s => !isOverallQuicklyIncreasingSummary(s));
+
+    expect(overallQuicklyIncreasingSummaries.length).toBe(1);
+    expect(overallQuicklyIncreasingSummaries.every(isValidSummary)).toBeTrue();
+    expect(otherSummaries.some(isValidSummary)).toBeFalse();
```
Rhombus33, line 53: What does this test for? Is there a summary that this dataset should never generate? I think it would be better to be specific, since this might not protect against a breakage and would be hard to debug the problem if it did break.

Here and in the similar expect case in the next test.

src/datasets/summarizations/trend-regression.summarization.spec.ts
Rhombus33: Added comments on new tests! Thanks for adding them and expanding test coverage 👍

src/datasets/summarizations/trend-regression.summarization.ts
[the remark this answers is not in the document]
Rhombus33: How would you describe them separately?

src/datasets/summarizations/trend-regression.summarization.ts
```diff
@@ -0,0 +1,154 @@
...
+          return 0;
+      }
+    };
+
+    const uWeekday = (p: TimeSeriesPoint) => 1 - uWeekend(p);
+
+    const isWeekend = (p: TimeSeriesPoint) => uWeekend(p) > 0.5;
+    const isWeekday = (p: TimeSeriesPoint) => uWeekday(p) > 0.5;
+
+    const weekdayWeekendRatioPoints = groupPointsByXWeek(points).map(weekPoints => {
```
Rhombus33, line 53: Could you add a description to this function and what it outputs?

src/datasets/summarizations/trend-regression.summarization.ts
```diff
@@ -0,0 +1,154 @@
...
+    const MAX_ANGLE = Math.atan(500 / 800);
+    const uQuicklyIncreasingLinearTrend = trapmfL(MAX_ANGLE / 2, MAX_ANGLE * 5 / 8);
+    const uIncreasingLinearTrend = trapmf(MAX_ANGLE / 8, MAX_ANGLE / 4, MAX_ANGLE / 2, MAX_ANGLE * 5 / 8);
+    const uConstantLinearTrend = trapmf(-MAX_ANGLE / 4, -MAX_ANGLE / 8, MAX_ANGLE / 8, MAX_ANGLE / 4);
+    const uDecreasingLinearTrend = trapmf(-MAX_ANGLE * 5 / 8, -MAX_ANGLE / 2, -MAX_ANGLE / 4, -MAX_ANGLE / 8);
+    const uQuicklyDecreasingLinearTrend = trapmfR(-MAX_ANGLE * 5 / 8, -MAX_ANGLE / 2);
+
+    const uSmallRegressionError = trapmfR(0.75, 1.0);
+
+    const uWeekend = (p: TimeSeriesPoint) => {
```
Rhombus33, line 35: probabilityDayIsWeekend better represents that this is a function

src/datasets/summarizations/trend-regression.summarization.ts
```diff
@@ -0,0 +1,154 @@
...
+          return 0.2;
+        case 6: // Saturday
+        case 0: // Sunday
+          return 1;
+        default: // All other days
+          return 0;
+      }
+    };
+
+    const uWeekday = (p: TimeSeriesPoint) => 1 - uWeekend(p);
```
Rhombus33, line 48: probabilityDayIsWeekday better represents that this is a function

src/datasets/summarizations/trend-regression.summarization.ts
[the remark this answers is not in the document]
Rhombus33: If a variable is hard to name, it's fine to add a comment for clarity as well.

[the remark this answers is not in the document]
jcc37: (math symbols) s, w
(correct but abstract) numerator and denominator 
(specific but confusing) weekday weighted y sum and weekday sum of weights

src/datasets/summarizations/trend-regression.summarization.ts
jcc37: This is a membership function on time series points. The membership functions in mathematics are usually denoted as "mu" function. That is why I named all the membership functions with "u" as the prefix.

https://en.wikipedia.org/wiki/Membership_function_(mathematics)

src/datasets/summarizations/trend-regression.summarization.spec.ts
jcc37: The summarization is designed to return all summary candidates with the difference in validity values, see the test of `should return all combinations of subsets and linguistic variables.`. So how I verify the returned summaries is by checking the validity of the "expected" summary should be high and others should be low.

Rhombus33: Do you expect them all to do low? If so can we use otherSummaries.every instead? More specific is better for testing.

src/datasets/dummy.dataset.ts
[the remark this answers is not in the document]
Rhombus33: A description in the UI too is a good idea!

src/datasets/dummy.dataset.ts
```diff
@@ -18,13 +19,17 @@ export const configMeta: PreferenceMeta<Config> = {
 };
 
 export function create(config: Config): Dataset {
-  const time = new Date(Date.now() - 100 * DAY);
+  const startDaysBeforeNow = 100;
+  const expContinuousGrowthRate = 0.1;
+
+  const time = new Date(Date.now() - startDaysBeforeNow * DAY);
```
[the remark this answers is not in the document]
Rhombus33: rename?

PR #83, opened 2020-08-05
Add y-movement and hover-design cards

cvillapudua:
- added `keyDown` event listener to document to register `q` and `z` as y-dimension controls
- added `daydream-control` with nested `raycaster` and `cursor` to act as a pointer in VR mode
- changed design of text/data cards to be a hover-card design (info displays upon hover of sphere) as opposed to previous design of distance ( PR #82 )


src/d3/scatterplot.d3.ts
```diff
@@ -44,14 +45,34 @@ export class Scatterplot{
...
       this.generateText();
+      this.generatePts();
       this.setColor('blue');
     }
     this.createSky('gray');
     this.createGridPlane();
     this.loaded = true;
+    window.onload = () => {
+      document.addEventListener('keydown', (event) => {
+        if (event.keyCode === 81){
```
mfitz3, line 57: I think keyCode is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Can use `code`: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

src/d3/scatterplot.d3.ts
```diff
@@ -4,8 +4,7 @@ import { VRScatterPoint } from '../datasets/queries/vr.query';
 import { MetaType } from '../datasets/metas/types';
 const PROXY_FLAG = '__keyboard-controls-proxy';
 const DATA_PT_RADIUS = .05;
-
-
+// const CAM_Y_ROTATION = AFRAME.THREE.MathUtils.degToRad(135);
```
mfitz3, line 7: Remove comment.

src/d3/scatterplot.d3.ts
```diff
@@ -124,9 +125,11 @@ export class Scatterplot{
   }
 
   private generateText(){
+    this.scalePosition();
     // enter identifies any DOM elements to be added when # array elements doesn't match
     d3.select(this.dataTextContainer).selectAll('a-entity').data(this.data).enter().append('a-entity');
-    d3.select(this.dataTextContainer).selectAll('a-entity')
+    this.cardSelection =  d3.select(this.dataTextContainer).selectAll('a-entity');
```
mfitz3, line 131: Do you need the variable? E.g.:

```
d3.select(this.dataTextContainer).selectAll('a-entity')
  .attr('geometry', 'primitive: plane; height: auto; width: .5')
  .attr(...);
...

cvillapudua: I end up needing this to filter through when the user is hovering over a sphere and wants to look at a card... I add a `mouseenter` event listener to the sphere that will use `cardSelection`

PR #84, opened 2020-08-05
Controller Interface Layout Update

HarrisonPearl:
Updates the layout of the scene to better facilitate controller interactions
- reduces size of the grids and points to fit in an approximately 2ft by 2ft by 2ft area in front of the user
- adds point size and color animations to indicate when haptics are being triggered, and their intensity
- adds scale functions to map ingested data to haptic intensities and position generation
- updates function descriptions and variable names for consistency + clarity
- adds dynamic position testing, making updating testing on layout changes more efficient

### DIFF
--- a/src/d3/hapticplot.d3.spec.ts
+++ b/src/d3/hapticplot.d3.spec.ts
@@ -1,3 +1,4 @@
+import * as d3 from 'd3';
 import { Hapticplot } from './hapticplot.d3';
 import { Entity, Scene } from 'aframe';
 import { Vector3 } from 'three';
@@ -11,6 +12,11 @@ describe('VR Haptic Plot', () => {
   let scene: HTMLElement;
   let controller: HTMLElement;
   let hapticplot: Hapticplot;
+  let graphScale: d3.ScaleLinear<number, number>;
+
+  const POINT_SIZE = 0.02;
+  const DEFAULT_COLOR = '#F0A202';
+  const HOVER_COLOR = 'red';
 
   beforeEach( () =>  {
     scene = document.createElement('a-scene');
@@ -22,6 +28,8 @@ describe('VR Haptic Plot', () => {
     (controller as Entity).sceneEl = (scene as Scene);
     scene.appendChild(controller);
     hapticplot = new Hapticplot(shape);
+
+    graphScale = d3.scaleLinear();
   });
 
   it('places no points bc 1:1 correspondence with empty data array', () => {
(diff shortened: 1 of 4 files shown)


src/d3/hapticplot.d3.ts
```diff
@@ -114,22 +123,22 @@ export class Hapticplot{
     const xGrid = document.createElement('a-entity');
     xGrid.id = 'xGrid';
     this.container!.appendChild(xGrid);
-    xGrid.object3D.add(new THREE.GridHelper(50, 50, 0xffffff, 0xffffff));
-    d3.select(this.container).select('#xGrid').attr('position', '0 0 0');
+    xGrid.object3D.add(new THREE.GridHelper(1, 50, 0xffffff, 0xffffff));
+    d3.select(this.container).select('#xGrid').attr('position', '0 1 -1');
     d3.select(this.container).select('#xGrid').attr('rotation', '0 0 0');
```
danielyule, line 128: Here (and elsewhere in the file), you can cascade these (put the `.attr`s right after each other instead of reselecting)

src/d3/hapticplot.d3.spec.ts
```diff
@@ -32,29 +41,39 @@ describe('VR Haptic Plot', () => {
...
   it('places points for each datum in a one datum array', () => {
-    hapticplot.init(scene, [10]);
-    const expectedPosArray = [new Vector3(0, 10, -1)];
+    const dataArray = [10];
+    graphScale.domain([0, d3.max(dataArray) as number])  // max of dataset
+      .range([0, 0.5]);
+    hapticplot.init(scene, dataArray);
+    const expectedPosArray = Array.from(dataArray).map(
+      (datum: number, i) =>
+      new Vector3((0.5 / dataArray.length) * i, graphScale(datum) + 1, -1));
```
danielyule, line 50: I understand the impulse to generate the expected array from the input array, because it makes more clear the connection between the input and expected values, and also it means that changing something in one place will automatically update it in another. I've done this in the past.

But in tests, it's usually better to have explicitly constructed expected values. The more logic you have in your test case, the more likely there's an error with your test, and not with the code you're testing; after all, you're not testing your test.  In this case in particular there is an even bigger issue, which is that the code you're using in your tests is exactly the same as the code in your class that you are trying to test.  So if you made an error in your logic in one place, it's going to show up in the other as well.

(side note): we have (had)? this thing at Google called "Testing on the Toilet" which puts a little piece of paper inside each toilet stall with some tips on good testing practices.  This exact topic was covered in one of the issues, so this is a very common pattern, even among regular Googlers.

src/d3/hapticplot.d3.ts
```diff
@@ -6,24 +6,28 @@ export class Hapticplot{
...
   constructor(shape: string) {
     this.shape = shape;
   }
 
   init(container: HTMLElement | null, data: number[]){
-    const POINT_SIZE = 0.05;
-    const DEFAULT_COLOR = 'green';
+    const POINT_SIZE = 0.02;
+    const DEFAULT_COLOR = '#F0A202';
     const HOVER_COLOR = 'red';
```
danielyule, line 19: Slight preference for these to be put outside the class (but in the file)

src/d3/hapticplot.d3.ts
```diff
@@ -59,52 +63,57 @@ export class Hapticplot{
...
+      .attr('radius', size);
   }
 
   // Creates and adds a sky box to the scene
   private createSky(){
     const aSky = document.createElement('a-sky');
     this.container!.appendChild(aSky);
     d3.select(this.container).selectAll('a-sky').attr('color', () => {
-      return '#f2b9af';
+      return '#4d4d4d';
```
danielyule, line 116: This could probably be a constant as well

PR #87, opened 2020-08-10
Adding audio

HarrisonPearl:
Adds audio feedback in the form of marimba notes. When users' controllers come in to contact with a data point, a marimba note is emitted with a tone mapped from the points data.
- includes mp3 files for 27 notes arranged by increasing pitch in CDEFGABC order between 4 octaves.
- sets up triggers on generated data points which map each of their relative values to 1 of the 27 notes, and plays the associated mp3 on hover events
- set up testing to insure generated objects have sound components and triggers properly attached


src/d3/hapticplot.d3.ts
```diff
@@ -83,7 +86,19 @@ export class Hapticplot{
   }
 
   /**
-   * Triggers a haptic pulse, changes a points color and size when a it is hovered by the controller entity
+   * Attaches audio triggers to each point, mapping associated data to mp3 marimba notes
+   * @param datum Ingested Data
+   * @param point the point who's position is being set
+   */
+  private setAudio(datum, point){
```
danielyule, line 93: I noticed while testing this that hitting the same point several times in a row doesn't replay the same noise unless something amount of time has elapsed.  Is this intentional?

src/components/vr-accessibility/vr-accessibility.component.ts
```diff
@@ -32,7 +32,7 @@ export class VRAccessibilityComponent implements OnInit, OnChanges, OnDestroy, A
 
   ngAfterViewInit(){
     const scene = this.theScene.nativeElement;
-    this.vrHapticPlot.init(scene, [0, 1, 4, 9, 16, 25, 36, 49, 64, 81, 100]);
+    this.vrHapticPlot.init(scene, [...Array(27).keys()]);
```
danielyule, line 35: I think it will be useful to try to integrate some randomly generated data here to see what it will look like in a more realistic use case.  Maybe in a follow up PR.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -93,6 +93,19 @@ describe('VR Haptic Plot', () => {
     const result = hasHaptics(controller);
     expect(result).toEqual(expectedRes);
   });
+
+  it('initilizes data points with sound components', () => {
+    hapticplot.init(scene, [10, 20, 30]);
+    const a = hasSounds(scene, shape);
+    expect(a).toEqual(true);
```
nattySP, line 100: nit: in this case you aren't gaining anything with this single letter intermediate variable;  consider replacing with 
```suggestion
    expect(hasSounds(scene, shape)).toEqual(true);
```
(same below).

more generally, variable names should tell you what they represent, and single letter variables are only typically used as iteration counters (e.g let i; i < bounds; i++); many teams opt to have their linters yell at you for single letter variables.

src/d3/hapticplot.d3.ts
```diff
@@ -83,7 +86,19 @@ export class Hapticplot{
...
+    d3.select(point)
+      .attr('sound', 'src: url(../../assets/'
+        + Math.round(this.audioScale(datum)).toString()
+        + '.mp3); on: hover-start');
```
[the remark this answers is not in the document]
HarrisonPearl: done! I switched to a temple, created a constant, and added a dependency to [https://www.npmjs.com/package/@braintree/sanitize-url](url)

src/d3/hapticplot.d3.ts
HarrisonPearl: It isn't. I think that is coming from the fact that the mp3 files have ~1 second a quite reverb at the end, and calling play on one of them while it is already playing won't interrupt it. I think a good work around is checking if a sound is playing before playing it, and stopping it if it is.

src/components/vr-accessibility/vr-accessibility.component.ts
HarrisonPearl: Agreed! This one is laid out to test each of the audio files.

src/d3/hapticplot.d3.spec.ts
HarrisonPearl: fixed it!

src/d3/hapticplot.d3.ts
```diff
@@ -21,22 +22,22 @@ export class Hapticplot{
...
+      .range([0, 27]);
     this.setupPoints(DEFAULT_COLOR, HOVER_COLOR, POINT_SIZE);
     this.createSky();
     this.createGridPlane();
   }
 
-  /**
-   * selects all entities of type datapoint
-   */
+  // Selects all entities of type datapoint
```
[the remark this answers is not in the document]
HarrisonPearl: Done, I also update the single line comments elsewhere.

PR #92, opened 2020-08-17
Haptic Guidance

HarrisonPearl:
This update adds Haptic Guidance for data point navigation, and support for repeat sound triggers.

Haptic:
- A custom a-frame component, `guide`,  is added, which attaches to each controller entity.
- Each time a new frame is rendered, the `guide` components calculates the appropriate haptic feedback for each controller based on the distance between each and the closest data point.
- As a controller nears a data point, the haptic intensity increases.
- This increase in intensity is based on a mapping from the controller's proximity to the point, to two quadratic functions: one which determines intensity growth at long range, and another that introduces rapid growth at near proximity. This combination method was based on a drop of in precision of users ability to perceive differences at high vibration intensities.  
- Once contact is made with a data point, the controller vibrations switch to interval pulses to indicate that a point is being touched.
- Two custom type interfaces are also included for the component and its state property: `Guidance` and `State`
- Testing for each case of haptic behavior is included.

Sound:
- A check is added to stop a sound if it is currently playing when triggered, allowing sounds to be restarted mid play through.
- This allows sounds to be played repeatedly when a single data point is collided with multiple times in quick succession.
- Which means you can play the graph like an instrument! 

Note: This update is focused on interface design, and implements a naive, memory intensive method of nearest data point selection. This can cause lag when large datasets are introduced. The following update will focus on improving the efficiency of this selection using the kd tree data structure.

### DIFF
--- a/src/aframe_components/custom-components.ts
+++ b/src/aframe_components/custom-components.ts
@@ -1,0 +1,87 @@
+import { Vector3 } from 'three';
+import { Component } from 'aframe';
+
+const HAPTIC_RANGE = 0.3;
+const COLLISION_DISTANCE = 0.029;
+const HAPTIC_EXPONENT = 2;
+const HAPTIC_GROWTH_MODIFIER = 4;
+const HAPTIC_DURATION = 5000;
+const INTERVAL_DURATION = 100;
+
+interface State {
+  currInterval: number;
+  currPosition: Vector3;
+  distances: Vector3[];
+  currDistance: number;
+  currDistanceScaled: number;
+  intensity: number;
+}
+
+export interface Guidance extends Component {
+  state: State;
+  pointPositions: Vector3[];
+}
+
+export const guideComponent: Partial<Guidance> = {
+  state: {
+    currInterval: INTERVAL_DURATION,
+    currPosition: new Vector3(0, 0, 0),
+    distances: [],
+    currDistance: HAPTIC_RANGE,
+    currDistanceScaled: (HAPTIC_RANGE - COLLISION_DISTANCE) * (HAPTIC_RANGE) / (HAPTIC_RANGE - COLLISION_DISTANCE),
+    intensity: 0,
+  },
+
+  pointPositions: [],
+
+  init(){
+    this.pointPositions =
+      Array.from((this.el.sceneEl.querySelectorAll('a-sphere') as Element[])).map((point: Element) =>
+      point.getAttribute('position'));
+    this.state.currInterval = INTERVAL_DURATION;
+  },
+
+  /**
+   * A function called each render frame which handles the controller haptics
+   * - increases haptic intensity as controller nears a datapoint within HAPTIC_RANGE
+   * - sends hapic pules on a repeating interval once contact is made with a point
+   * @param time global scene uptime
+   * @param timeDelta time elapsed since previous frame
+   */
+  tick(time, timeDelta) {
+    this.state.currPosition = this.el.object3D.position;
+    this.state.distances = Array.from(this.pointPositions as Vector3[]).map((pos: Vector3) => this.state.currPosition.distanceTo(pos));
+    this.state.distances.sort();
+    this.state.currDistance = this.state.distances[0];
+    // If a datapoint is within haptic range, send haptic feedback based on its proximity
+    // - uses a combination of quadratic functions to map proximity to haptic intensity
+    if (COLLISION_DISTANCE < this.state.currDistance && this.state.currDistance < HAPTIC_RANGE){
+      // Translates distance from COLLISION_DISTANCE -> HAPTIC_RANGE to 0 -> HAPTIC_RANGE, for use in intensity calculations
+      this.state.currDistanceScaled = (this.state.currDistance - COLLISION_DISTANCE) * (HAPTIC_RANGE) / (HAPTIC_RANGE - COLLISION_DISTANCE);
+      this.state.intensity = Math.max(
+        Math.pow((this.state.currDistanceScaled - HAPTIC_RANGE) / (HAPTIC_RANGE * 2), HAPTIC_EXPONENT),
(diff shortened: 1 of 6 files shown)


src/d3/hapticplot.d3.spec.ts
```diff
@@ -177,3 +220,46 @@ function hasSoundTriggers(scene: HTMLElement, shape: string): boolean{
 function hasGuidance(controller: HTMLElement): boolean{
   return (controller as Entity).components.hasOwnProperty('guide');
 }
+
+// Initilizes the guide component, and returns the currInterval
+function getGuidanceInterval(controller: HTMLElement): number{
+  const controllerEntity = controller as Entity;
+  controllerEntity.components.guide .init();
```
danielyule, line 227: extra space

src/d3/hapticplot.d3.spec.ts
```diff
@@ -177,3 +220,46 @@ function hasSoundTriggers(scene: HTMLElement, shape: string): boolean{
...
+}
+
+// Returns the current inteval value after a tick of duration 100ms
+function getIntervalReset(controller: HTMLElement): number{
+  const controllerEntity = controller as Entity;
+  controllerEntity.object3D.position.set(0, 1.7, -0.35);
+  controllerEntity.components.guide.init();
+  controllerEntity.components.guide.tick!(100, 100);
+  return (controllerEntity.components.guide as Guidance).state.currInterval;
+}
```
danielyule, line 265: In a future PR, you may want to consider moving all these utility functions into their own file.

src/aframe_components/custom-components.ts
```diff
@@ -0,0 +1,87 @@
...
+  currDistanceScaled: number;
+  intensity: number;
+}
+
+export interface Guidance extends Component {
+  state: State;
+  pointPositions: Vector3[];
+}
+
+export const guideComponent = {
```
danielyule, line 25: `export const guideComponent: Guidance = {`
I think this will mean less casting when you access `state` or `pointPositions`

src/aframe_components/custom-components.ts
```diff
@@ -0,0 +1,87 @@
...
+    currDistance: HAPTIC_RANGE,
+    currDistanceScaled: (HAPTIC_RANGE - COLLISION_DISTANCE) * (HAPTIC_RANGE) / (HAPTIC_RANGE - COLLISION_DISTANCE),
+    intensity: 0,
+  },
+
+  pointPositions: [],
+
+  init(){
+    this.pointPositions =
+      Array.from(this.el.sceneEl.querySelectorAll('a-sphere')).map((point: unknown) =>
```
danielyule, line 39: could it be `point: Element`? Then you wouldn't need the cast on the next line.

src/aframe_components/custom-components.ts
```diff
@@ -0,0 +1,87 @@
...
+   * A function called each render frame which handles the controller haptics
+   * - increases haptic intensity as controller nears a datapoint within HAPTIC_RANGE
+   * - sends hapic pules on a repeating interval once contact is made with a point
+   * @param time global scene uptime
+   * @param timeDelta time elapsed since previous frame
+   */
+  tick(time, timeDelta) {
+    this.state.currPosition = this.el.object3D.position;
+    this.state.distances = Array.from(this.pointPositions as Vector3[]).map((pos: Vector3) => this.state.currPosition.distanceTo(pos));
+    this.state.distances.sort();
```
danielyule, line 54: I know this is temporary, but any gains made from not declaring local variables is definitely wiped out by creating two arrays (Array.from, and map) and then sorting them.  There's nothing to change, but it's important to consider where the bulk of the time will actually be spent, and if things like maintaining the state like you have here will actually lead to performance gains.

In this particular case it's fine because it sounds like predeclaring things is standard AFrame practice, and you generally want to follow standard practice whenever possible.

src/aframe_components/custom-components.ts
```diff
@@ -0,0 +1,53 @@
...
+const COLLISION_DISTANCE = 0.029;
+const HAPTIC_EXPONENT = 2;
+const HAPTIC_GROWTH_MODIFIER = 4;
+const HAPTIC_DURATION = 5000;
+const INTERVAL_DURATION = 100;
+
+export const guideComponent = {
+    init(){
+      this.pointPositions =
+        Array.from(this.el.sceneEl.querySelectorAll('a-sphere')).map((point: unknown) =>
```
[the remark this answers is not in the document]
HarrisonPearl: Yes! But doing so also requires casting the result of `querySelectorAll` to an `Element[]` I've added both, and removed `unknown`

src/aframe_components/custom-components.ts
HarrisonPearl: I changed to exporting with the Guidance type, which required adding the properties of the component type (which my guidance interface extends) to the guidance component definition.

I did this by including the properties of the component as follows:
el: this.el,
data: this.data

In order to maintain the components default values while including an explicit definition of them in the guidance component. Is this a good way to handle this?

HarrisonPearl: I tried this, but I think guideComponent isn't of type Guidance. It is used as the `description` property when I call AFRAME.registerComponent in maine.ts and test.ts. registerComponent return the component, which is of type guidance. But, when I'm accessing the component using entity.components.guide in my testing, it is still being returned as a generic component, which I then cast as type Guidance in order to access its properties. I'm not sure how to get around this.

src/aframe_components/custom-components.ts
HarrisonPearl: This is definitely a performance hog! I'm using it as a place holder so I could test out the interface before working on the kd tree.

I was messing around with rendering large numbers of variables to check out the effect of this naive selection on the render cycle. Without the guidance component, I can render around a thousand data points before fps starts to drop significantly. Enabling guidance with this type of selection causes major lag with around 300 points. Good proof of concept for the benefits of implementing the kd tree!

PR #94, opened 2020-08-20
Weekly comparison - rate text summarization

jcc37:
Add `TrendWeeklyComparisonRate` summarization module and tests for time-series data.
The summaries describe the comparison between the linear increasing/decreasing rate of y-values of every two consecutive weeks. The following are sample summaries for a one-month time-series data:
* The active users of weekdays in the second week increased in the same rate as the first week.
* The active users of weekdays in the third week increased 31.3% (4.2 less user increased per day) slower than the second week.
* The active users of weekdays in the fourth week increased 61.9% (5.7 more user increased per day) faster than the third week.


src/datasets/summarizations/trend-weekly-comparison-rate.summarization.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  it('should create summaries describing monotonic linearly increasing data.', () => {
+    const points: TimeSeriesPoint[] = [];
+    for (let i = 6; i <= 26; i++) {
+      const x = new Date(2020, 6, i);
+      const y = i * 10;
+      points.push({ x, y });
+    }
+
+    const isFirstWeekSecondWeekSummary = isTextPartsInSummary('second week', 'first week', 'increased', 'in the same rate');
```
Rhombus33, line 30: Just curious, could we match a more complete sentence for these tests instead of 4 substrings?

Rhombus33 (approved):

src/datasets/summarizations/trend-weekly-comparison-rate.summarization.spec.ts
jcc37: I want the test to be robust to small changes in summary templates. Also there might be some numerical values in the summaries.

Rhombus33: My only concern is that depending on what other words are in the summary, the meaning could be very different. I think that there are not so many tests that changing them each time the template changes would be too much of a hassle.

PR #98, opened 2020-08-27
Nearest Point

HarrisonPearl:
Adds efficient nearest neighbor point selection using a[ k-d tree ](https://en.wikipedia.org/wiki/K-d_tree) data structure.
- On each frame update, each controller must determine its haptic behavior based on its proximity to the closest data point.
- This PR adds an efficient means of doing so by:
  - building a 3-dimensional k-d tree of the data points' during the initial scene set up.
  - using the tree for selection of the nearest neighbor to the controllers' positions each frame, with time complexity O(3logN) in relation to the number of data points.
- The guide component is attached to a controller in order to enable nearest neighbor selection and haptic feedback.

### DIFF
--- a/src/aframe_components/custom-components.ts
+++ b/src/aframe_components/custom-components.ts
@@ -1,5 +1,6 @@
 import { Vector3 } from 'three';
 import { Component } from 'aframe';
+import { build3DTree, distToNearestPoint, KDNode} from './kdtree';
 
 const HAPTIC_RANGE = 0.3;
 const COLLISION_DISTANCE = 0.029;
@@ -11,7 +12,6 @@ const INTERVAL_DURATION = 100;
 interface State {
   currInterval: number;
   currPosition: Vector3;
-  distances: Vector3[];
   currDistance: number;
   currDistanceScaled: number;
   intensity: number;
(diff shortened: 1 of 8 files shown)


src/aframe_components/custom-components.ts
```diff
@@ -20,24 +25,100 @@ interface State {
...
+
+  const distanceA: number = controllerPos.distanceTo(pointA);
+  const distanceB: number = controllerPos.distanceTo(pointB);
+
+  if (distanceA < distanceB){
+    return pointA;
+  }
+  else{
+    return pointB;
+  }
```
danielyule, line 101: Probably a good idea to move the tree node implementation into its own file.  Then it can be tested independently, and can also be used elsewhere (theoretically)

src/aframe_components/custom-components.ts
```diff
@@ -20,24 +25,100 @@ interface State {
...
+    nextNode = node.right;
+    otherNode = node.left;
+  }
+
+  let closest = closerDistance(
+    controllerPos,
+    node.position,
+    nearestPoint(controllerPos, nextNode, depth + 1));
+
+  if (controllerPos.distanceTo(closest!) > Math.abs(controllerPos[AXIS_DICT[axis]] - node.position[AXIS_DICT[axis]])){
```
danielyule, line 75: Here and everywhere, use distanceToSquared, as it's faster to compute.

src/aframe_components/custom-components.ts
```diff
@@ -20,24 +25,100 @@ interface State {
...
+    closest = closerDistance(
+      controllerPos,
+      closest,
+      nearestPoint(controllerPos, otherNode, depth + 1));
+  }
+
+  return closest;
+}
+
+function closerDistance(controllerPos: Vector3, pointA: Vector3 | null, pointB: Vector3 | null){
```
danielyule, line 85: If you return both the point and the distance, then you won't have to recompute the distance when doing the "unwinding" step.

src/aframe_components/custom-components.ts
```diff
@@ -20,24 +25,100 @@ interface State {
...
+    position: positions[mid],
+    left: build3DTree(positions.slice(0, mid), depth + 1),
+    right: build3DTree(positions.slice(mid + 1, n), depth + 1),
+  };
+}
+
+export function nearestPoint(controllerPos: Vector3, node: KDNode | null, depth: number){
+  if (node == null){
+    return null;
+  }
```
danielyule, line 55: Here and elsewhere, slight preference for `if (!node) {`

src/d3/hapticplot.d3.spec.ts
```diff
@@ -140,6 +137,33 @@ describe('VR Haptic Plot', () => {
...
+        },
+        right: null,
+      },
+      right: {
+        position: new Vector3( 0.5249999999999999, 1.7, -0.35),
+        left: null,
+        right: null,
+      },
+    });
+  });
```
danielyule, line 160: I think when you move the KD tree into its own file, you may want a little more comprehensive testing, making sure that it works in some degenerative cases (like all the points are in a vertical line, or just a single point, or if the controller's position exactly equals one of the points in the data array.

This is good both as a general practice, but also in this particular case to make it easier to refactor the tree in case you want to optimize it in some way later on.  You can be more confident you didn't break anything with a bunch of tests.

src/aframe_components/custom-components.ts
```diff
@@ -20,24 +25,100 @@ interface State {
...
   state: State;
   pointPositions: Vector3[];
+  kdtree: KDNode | null;
+}
+
+export interface KDNode {
+  position: Vector3;
+  left: KDNode | null;
+  right: KDNode | null;
+}
```
danielyule, line 35: Not a request for a change, just curious if you gave any consideration to implementing this as a class instead of an interface with associated functions.

src/aframe_components/custom-components.ts
```diff
@@ -20,24 +25,100 @@ interface State {
...
+
+export function build3DTree(positions: Vector3[], depth: number){
+  const n = positions.length;
+  const mid = Math.floor(n / 2);
+  if (n <= 0){
+    return null;
+  }
+  const axis = depth % 3;
+  positions.sort((a, b) => a[AXIS_DICT[axis]] - b[AXIS_DICT[axis]]);
```
danielyule, line 44: Did you give any thought to presorting this array by each dimension ahead of time?

src/aframe_components/custom-components.ts
```diff
@@ -20,24 +25,100 @@ interface State {
...
+  return {
+    position: positions[mid],
+    left: build3DTree(positions.slice(0, mid), depth + 1),
```
danielyule, line 47: Instead of slicing the array at every step, you could pass a start and end index into `build3DTree` and operate on only those.  This way you're not allocating two arrays each time you recurse

src/aframe_components/custom-components.ts
```diff
@@ -50,36 +52,41 @@ export const guideComponent: Partial<Guidance> = {
    */
   tick(time, timeDelta) {
     this.state.currPosition = this.el.object3D.position;
```
nattySP, line 54: Have you considered, instead of mutating the state, building a new state object in this function and assigning it at the end to this.state?  I get nervous when I see a lot of object mutations because it can lead to very difficult to track bugs, especially if you are mutating from different methods etc.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -140,6 +137,101 @@ describe('VR Haptic Plot', () => {
...
+    expect(getDistToNearestPoint(controller, scene, shape, new Vector3(0, 0, 0))).toEqual(1.735655495770978);
+  });
+
+  it('correctly identifies the nearest data point when its positon is the same as the controllers', () => {
+    const dataArray = [10];
+    hapticplot.init(scene, dataArray);
+    expect(getDistToNearestPoint(controller, scene, shape, new Vector3(0, 1.7, -0.35))).toEqual(0);
+  });
+
+  it('correctly identifies the nearest data point when each point has the same value', () => {
```
nattySP, line 217: which point is considered the nearest in this case? consider adding that detail to your expectation.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -263,3 +355,35 @@ function getIntervalReset(controller: HTMLElement): number{
   controllerEntity.components.guide.tick!(100, 100);
   return (controllerEntity.components.guide as Guidance).state.currInterval;
 }
+
+// Returns the current inteval value after a tick of duration 100ms
```
nattySP, line 359: typo: interval

src/aframe_components/custom-components.ts
danielyule: I think ours ONLY lets it through if it's with the null, like this one.

src/d3/hapticplot.d3.spec.ts
```diff
@@ -140,6 +137,131 @@ describe('VR Haptic Plot', () => {
...
+        },
+        right: null,
+      },
+      right: {
+        position: new Vector3(0.5249999999999999, 1.7, -0.35),
+        left: null,
+        right: null,
+      },
+    });
+  });
```
danielyule, line 227: All the stuff about building and probing the kd-tree should be in a separate test spec file.  Then if you need to move the kd stuff elsewhere, it's obvious what needs to go.  It also makes for better decoupling.

src/aframe_components/custom-components.ts
```diff
@@ -50,36 +52,41 @@ export const guideComponent: Partial<Guidance> = {
...
+        if (this.state.currInterval > INTERVAL_DURATION / 2){
+          this.state.intensity = 1;
+        }
+        else {
+          this.state.intensity = 0;
+        }
+
+        if (this.state.currInterval <= 0){
+          this.state.currInterval = INTERVAL_DURATION;
+        }
```
danielyule, line 83: Consider extracting some of this to another function.

danielyule (approved):

src/aframe_components/custom-components.ts
HarrisonPearl: I was under the impression that using an interface was a bit lighter on overhead, and so more efficient, but that a class organization may be helpful for readability. Would you recommend one over the other?

src/d3/hapticplot.d3.spec.ts
HarrisonPearl: I only return the closest distance while searching my tree, and use that value as the input to call my closerPoint and closerPointUnwind search functions recursively. I can add logic to make sure the point is also passed forward through these, but this may have an efficiency effect, and I don't need it for my haptic calculations. I'm not sure how to access this in the testing without doing so. It may be worth doing for the generalizability of the function later on though?

PR #100, opened 2020-09-11
Text to speech

HarrisonPearl:
Adds spoken word readouts for data point positions
- Users can have the position of a data point read out to them by pushing the A or X button while the controller is touching the point
- Spoken Word audio is generated by playing a set of 15 audio files
  - the digits 0-9
  - the letters X, Y, and Z
  - the words "negative" and "point"
- On a button press a string is generated from the data points position in the format: X0.00Y0.00Z0.00
- This string is used to determine the order in which the audio files are played
- Each time a digit is spoken, a trigger is set up to play the next digit when the audio ends with a `sound-ended` event


src/d3/hapticplot.d3.ts
```diff
@@ -108,12 +105,54 @@ export class Hapticplot{
...
+    .on('xbuttonup',  (d, i, g) => this.speakPosition(datum, point));
+  }
+
+  /**
+   * Reads out the position of the given point in spoken audio, and resets its sound component to
+   * the appropriate marimba note audio upon completion
+   * @param datum The datum from which the given points position and audio is generated
+   * @param point A data point entity
+   */
+  private speakPosition(datum, point){
```
danielyule, line 125: I've just noticed that a bunch of private methods in this file don't have any types in the params.  I actually thought TS required that, so I wasn't checking, but they should all have types.

src/d3/hapticplot.d3.ts
```diff
@@ -108,12 +105,54 @@ export class Hapticplot{
...
+    const position = (d3.select(point).attr('position') as unknown) as Vector3;
+    const posString =
+      `X${Math.round((position.x - GRAPH_OFFSET.x) * 100) / 100}Y${Math.round((position.y - GRAPH_OFFSET.y) * 100) / 100}Z${Math.round((position.z - GRAPH_OFFSET.z) * 100) / 100}`;
```
danielyule, line 128: Any way to break this line up so it's not so long?

danielyule (approved):

src/d3/hapticplot.d3.ts
HarrisonPearl: I added them! Doing so revealed a secret any cast when I was accessing the sound component, so I added an interface for the Sound type extending the generic Component type

src/d3/hapticplot.d3.ts
HarrisonPearl: yes! I split them into separate literals for each axis and then concatenate them.

